PR #136, opened 2023-08-13
Added kernel SME filter

FlorianPfaff:


FlorianPfaff:
Need to check formula

PR #137, opened 2023-08-16
Reenabled grype now that false alarm was fixed

FlorianPfaff:


FlorianPfaff:
Probably needs new version of megalinter for new data base.

PR #198, opened 2023-09-14
Reordered some code for GaussianDistribution

FlorianPfaff:


FlorianPfaff:
Does not have a lot of merit

PR #200, opened 2023-09-14
Added beartype to AbstractParticleFilter

FlorianPfaff:


FlorianPfaff:
Name should be changed

PR #233, opened 2023-09-15
AbstractCustomDistribution: pdf now checks input dim

FlorianPfaff:


FlorianPfaff:
Might be solved by #1093

PR #422, opened 2023-12-01
Hypersphere: support conventions + explicit integrand fun

FlorianPfaff:


FlorianPfaff:
Features have been integrated via PRs #487 and #502

PR #573, opened 2024-02-03
Fixing error with plotting for latest dependencies

FlorianPfaff:


FlorianPfaff:
Did it in separate PR (#575) before asking dependabot to rebase

PR #861, opened 2024-10-02
Require numpy 2.1 or later

FlorianPfaff:


FlorianPfaff:
Maybe better stay flexible.

PR #899, opened 2024-10-30
Require jaxlib 0.4.33 or newer

FlorianPfaff:


FlorianPfaff:
Not a focus and the real minimum version has not been properly tested.

PR #1379, opened 2025-11-29
Also test Python 3.14

FlorianPfaff:
I see no reason why it will not work. Unavailable packages may prevent it from running for now, though.


FlorianPfaff:
May be built against other version of libhealpix.

PR #1405, opened 2025-12-02
Fixed regular expression in megalinter for pylint

FlorianPfaff:


FlorianPfaff:
Is not the cause of the problem.

PR #1663, opened 2026-04-09
Imported HypersphericalParticleFilter to __init__ notebook

Raghav030:
Imported HypersphericalParticleFilter to __init__ notebook so that it gets installed with main library


FlorianPfaff:
@Raghav030 Hi and thanks for your PR. However, can you check if this is actually still an issue with the latest version on GitHub?

PR #1873, opened 2026-04-25
Use Poetry for devcontainer setup

FlorianPfaff:
## Summary
- update the devcontainer image from Python 3.11 to Python 3.13, which is already covered by the test matrix
- replace the requirements-dev.txt pip install with a post-create script that installs the same system build dependencies used by CI and sets up dependencies with Poetry
- install Poetry 2.x plus the export plugin so the devcontainer matches the Poetry-based workflows

## Validation
- compared `codex/devcontainer-poetry-setup` against `main`; the branch is ahead by only the devcontainer JSON update and new post-create script
- did not run a full devcontainer build in this environment


FlorianPfaff:
Script not launched in codespaces

PR #2075, opened 2026-05-18
Fix von Mises moment inversion edge cases

FlorianPfaff:
## Summary
- replace the unconstrained `fsolve`-based Bessel-ratio inverse with a bracketed scalar solve
- validate first trigonometric moment magnitudes before moment matching
- handle the zero-moment uniform case explicitly and reject degenerate/infeasible moments
- add regression tests for moment reconstruction and invalid boundary values

## Notes
I did not run the test suite locally in this environment. Suggested check:

```bash
python -m pytest tests/distributions/test_von_mises_distribution.py
```


FlorianPfaff:
@copilot Fix merge conflicts

PR #2079, opened 2026-05-18
Fix JAX wrapped-normal PDF convergence check

FlorianPfaff:
This fixes the JAX branch of `WrappedNormalDistribution.pdf()` so the loop convergence check depends on the next wrapped addendum instead of the accumulated result.

Previously, the JAX loop condition effectively tested whether the accumulated density was positive. This could both run to `max_iterations` unnecessarily and terminate before adding important wrapped terms when the central term underflowed near the `0 / 2π` boundary.

Changes:
- factor out the JAX wrapped-term computation into `wrapped_addendum(i)`;
- continue while the next pair of wrapped terms changes at least one accumulated density value;
- add a regression test for periodicity when the central term underflows near the boundary.


FlorianPfaff:
@copilot Fix merge conflict

PR #2080, opened 2026-05-18
Handle undefined hyperspherical mean directions

FlorianPfaff:
Summary:
- Raise ValueError when a numerical first moment is effectively zero instead of normalizing it.
- Reuse the same normalization guard in the hyperhemispherical numerical mean-direction path.
- Add a regression test for the full-sphere uniform distribution, whose mean direction is undefined.

Testing: Not run locally because the changes were made through the GitHub connector without a local checkout.


FlorianPfaff:
@copilot Fix merge conflicts and CI/CD issues

PR #2231, opened 2026-05-25
Fix PyTorch sum keepdims contract

FlorianPfaff:
## Summary
- wrap the PyTorch backend `sum` export with a NumPy-compatible signature
- preserve `axis=None, keepdims=True` by reducing first and reshaping to singleton dimensions
- add backend-contract regression tests, including a subprocess test that forces `PYRECEST_BACKEND=pytorch`

## Rationale
The PyTorch backend implementation accepted `keepdims` but, for `axis=None`, forwarded it as `torch.sum(x, keepdim=...)`. PyTorch rejects `keepdim` unless a reduction dimension is supplied, while NumPy-compatible code expects `sum(values, keepdims=True)` to preserve singleton dimensions.

## Validation
- local PyTorch smoke check confirmed `torch.sum(x, keepdim=True)` raises the failing signature and that reducing first then reshaping reproduces NumPy-style output
- connector compare after rebasing shows this branch is ahead of current `main` by 2 commits and behind by 0
- changed files are limited to `src/pyrecest/_backend/__init__.py` and `tests/test_backend_sum_contract.py`
- full pytest was not run locally because the execution container cannot resolve `github.com`; PR CI should run the full backend matrix

### DIFF
--- a/src/pyrecest/_backend/__init__.py
+++ b/src/pyrecest/_backend/__init__.py
@@ -349,6 +349,27 @@ def _mean_with_numpy_signature(mean_func, asarray_func, reshape_func):
     return mean
 
 
+def _sum_with_numpy_signature(sum_func, asarray_func, reshape_func):
+    """Return a NumPy-compatible sum wrapper for stricter backends."""
+
+    @wraps(sum_func)
+    def sum(a, axis=None, dtype=None, out=None, keepdims=False):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as a duplicate of #2234, which carries the same PyTorch `sum(..., keepdims=True)` backend-contract fix and has auto-merge enabled.

PR #2989, opened 2026-06-24
Fix backend array-like helper contracts

FlorianPfaff:
Summary:
- Normalize array-like inputs in shared NumPy/autograd and JAX backend helpers.
- Make empty sparse support handling consistent across backends.
- Preserve JAX one-hot list label support with uint8 indicators.

Validation:
- compileall for backend modules
- ruff F821/F822/F823 checks on changed files
- targeted NumPy, PyTorch, and JAX regression checks for dimension, squeeze, sparse, and one-hot behavior


FlorianPfaff:

PR #3001, opened 2026-06-24
Treat boolean track matrix cells as missing

FlorianPfaff:
## Summary
- Treat boolean-like track matrix cells as missing values instead of observation IDs `0` and `1`.
- Add regression tests for boolean cells and non-scalar array cells.

## Testing
- CI should run on this PR.


FlorianPfaff:

PR #3015, opened 2026-06-24
Reject nonnumeric weak measurement covariance inputs

FlorianPfaff:
## Summary
- Validate weak-measurement standard deviation and covariance arrays before float coercion.
- Preserve boolean and text-like values as invalid configuration inputs instead of silently mapping them to numeric covariances.
- Add focused regressions for diagonal, block-diagonal, and direct covariance constructor paths.

## Bug
`src/pyrecest/models/weak_measurement.py` converted standard deviations and measurement covariance arrays with `np.asarray(..., dtype=float)` or `float(value)`. That silently accepted malformed inputs such as `True` or `"2.0"`, turning them into ordinary numeric standard deviations/covariances and hiding configuration/data errors.

## Validation
- Inspected the branch diff with GitHub compare: only `src/pyrecest/models/weak_measurement.py` and `tests/test_weak_measurement_input_validation.py` changed.
- Full pytest was not run locally because this environment cannot clone/install the repository from GitHub; CI should run the new focused regression tests on the PR.


FlorianPfaff:

PR #3037, opened 2026-06-25
NumPy zeros dtype fix

FlorianPfaff:
Preserve explicit positional dtype in the NumPy backend zeros wrapper and add regression coverage.


FlorianPfaff:

PR #3039, opened 2026-06-25
Preserve explicit positional dynamic dtypes

FlorianPfaff:
## Summary
- fix the shared `_dyn_update_dtype` wrapper so an explicitly supplied positional dtype is preserved instead of overwritten with the backend default
- add regression coverage for NumPy backend `zeros`, `empty`, and `linspace`

## Why
`_dyn_update_dtype` previously replaced the positional dtype argument whenever it was present. Calls such as `numpy_backend.empty((2,), np.float32)` or `numpy_backend.linspace(..., np.float32)` therefore silently returned the configured default dtype instead of the caller-requested dtype.

## Validation
- connector diff check: final branch diff is limited to the shared dtype-wrapper fix plus the new regression test file
- local standalone smoke check of the wrapper logic against NumPy passed
- full repository test suite not run here; the execution sandbox cannot resolve `github.com`, so I could not clone/install the repo in this environment


FlorianPfaff:
Verified the source fix locally. The failing PyTorch jobs come from the new NumPy-specific default-dtype test using the process-wide Torch default dtype. Isolating that test with a NumPy dtype makes the focused suite pass under NumPy, PyTorch, and JAX. This branch is also 125 commits behind main and should be refreshed before merge.

PR #3040, opened 2026-06-25
Reject nonnumeric pairwise covariance inputs

FlorianPfaff:
## Summary
- reject numeric-looking text, bytes, booleans, complex arrays, and object-dtype arrays before pairwise covariance helpers coerce inputs to float64
- preserve existing finite and shape validation after safe numeric coercion
- add regression coverage for invalid regularization/epsilon controls, boolean/text mean inputs, and boolean/text covariance stacks

## Testing
- `python -m py_compile /mnt/data/pairwise_covariance_features.py /mnt/data/test_pairwise_covariance_features.py`
- Not run: full pytest suite. The execution environment cannot clone/install the repository from GitHub because `github.com` DNS resolution is unavailable here.

### DIFF
--- a/src/pyrecest/utils/pairwise_covariance_features.py
+++ b/src/pyrecest/utils/pairwise_covariance_features.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 import math
 from typing import Any
 
+import numpy as _np
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3043, opened 2026-06-25
Reject boolean distribution weights

FlorianPfaff:
## Summary
- Reject boolean Dirac weights and boolean Dirac reweighting outputs.
- Reject boolean mixture weights before normalization.
- Add focused regression tests for Dirac and linear mixture weight validation.

## Testing
- Not run locally because the execution sandbox cannot resolve github.com.
- Added CI regression coverage in the touched test files.


FlorianPfaff:

PR #3050, opened 2026-06-25
Reject complex assignment costs

FlorianPfaff:
## Summary
- Reject complex-valued assignment cost matrices before NumPy can discard imaginary parts during float coercion.
- Reject complex row/column non-assignment costs with the same real-valued cost contract.
- Add regression coverage for complex dtype and object-dtype complex inputs.
- Refreshed the branch onto current `main` so the PR is mergeable again.

## Bug
`murty_k_best_assignments` and `min_cost_max_cardinality_assignment` validate assignment costs as real-valued costs, but complex NumPy arrays can be cast to `float` by discarding the imaginary component. That lets invalid model/output costs silently change the optimization objective. The new guards reject complex dtype and object-dtype complex values before coercion.

## Validation
- Not run locally in this connector environment.
- Added focused regression coverage in `tests/utils/test_assignment_complex_cost_validation.py`; CI should run the repository test matrix.

### DIFF
--- a/src/pyrecest/utils/assignment.py
+++ b/src/pyrecest/utils/assignment.py
@@ -26,6 +26,7 @@ from scipy.optimize import linear_sum_assignment
 
 _TEXT_TYPES = (str, bytes, bytearray, _np.str_, _np.bytes_)
 _BOOLEAN_TYPES = (bool, _np.bool_)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3056, opened 2026-06-25
Reject non-real block particle filter inputs

FlorianPfaff:
## Summary
- reject boolean, text-like, complex, datetime/timedelta, and other non-real numeric values before block particle filter weights, likelihoods, log-likelihoods, and ESS thresholds are coerced to floats
- preserve existing finite/nonnegative checks for valid real numeric arrays
- add focused regression coverage for malformed weights, likelihoods, and ESS thresholds

## Bug
`BlockParticleFilter` used direct `array(..., dtype=float)` conversion in several public update and weight-normalization paths. That allowed boolean masks such as `[True, False]` or numeric-looking text such as `["1.0", "0.0"]` to be interpreted as real weights/likelihoods/thresholds instead of surfacing malformed caller input.

## Validation
- `python -m py_compile /tmp/block_particle_filter.py`
- `python -m py_compile /tmp/test_block_particle_filter.py`

Full repository tests were not run locally because this environment cannot resolve github.com; CI should run the focused block-particle filter tests and the full matrix.

### DIFF
--- a/src/pyrecest/filters/block_particle_filter.py
+++ b/src/pyrecest/filters/block_particle_filter.py
@@ -57,6 +57,9 @@ def _as_integer(value, name: str, minimum: int | None = None) -> int:
 def _as_real_numeric_array(value, name: str):
     """Return ``value`` as a backend float array without silent nonnumeric coercion."""
 
+    if _has_unsupported_numeric_leaf(value):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3123, opened 2026-06-25
Reject temporal candidate pruning matrices

FlorianPfaff:
## Summary
- Reject temporal NumPy matrices before candidate-pruning helpers coerce inputs to float.
- Add regression coverage for temporal cost and probability matrices.

## Testing
- Regression test added in tests/test_candidate_pruning_temporal_validation.py.
- Not run locally because this environment cannot clone GitHub directly; CI should run on the PR.


FlorianPfaff:
Closing this duplicate/stale PR: the temporal candidate-pruning validation fix and regression test are already present on `main`.

PR #3160, opened 2026-06-26
Handle dtype-like values in shared default dtype setup

FlorianPfaff:
## Summary
- Normalize resolved backend dtype objects before deriving the paired complex dtype in the shared default-dtype setter.
- Avoid mutating the global dtype state when a dtype resolves to a non-floating type such as `complex64`.
- Add NumPy backend regression tests for dtype-like inputs and invalid non-floating dtype values.

## Bug
The shared `_pre_set_default_dtype` helper used the raw user-supplied `value` as the key for `_MAP_FLOAT_TO_COMPLEX` after resolving `DEFAULT_DTYPE`. For dtype-like inputs accepted by the NumPy backend, such as `np.dtype("float32")` or `np.float64`, the key lookup could fail even though the dtype itself resolved correctly. Non-floating dtype inputs could also partially mutate `DEFAULT_DTYPE` before failing while setting `DEFAULT_COMPLEX_DTYPE`.

## Testing
- `python -m py_compile /mnt/data/_dtype_utils.py /mnt/data/test_numpy_default_dtype_objects.py`
- Minimal local harness for `_pre_set_default_dtype(np.dtype)` covering `np.dtype("float32")`, `np.float64`, and rejected `np.dtype("complex64")` without state mutation.

Full repository tests were not run locally because this environment cannot clone/install the repository from GitHub.

### DIFF
--- a/src/pyrecest/_backend/_dtype_utils.py
+++ b/src/pyrecest/_backend/_dtype_utils.py
@@ -153,6 +153,17 @@ def _dyn_update_dtype(dtype_pos=None, target=None):
     return _decorator(target)
 
 
+def _dtype_name(value):
+    """Return a stable dtype name for backend dtype-like objects."""
+    name = getattr(value, "name", None) or getattr(value, "__name__", None)
+    if name is not None:
+        return str(name)
+    text = str(value)
+    if "." in text:
+        text = text.rsplit(".", maxsplit=1)[-1]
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded by newer default-dtype PRs and now conflicting with current main.

PR #3193, opened 2026-06-26
Fix transition sampler count handling

FlorianPfaff:
Summary:
- Preserve unary transition samplers for n=1.
- Raise ValueError when n is not 1 and the callback has no sample-count argument.
- Add regression tests for SampleableTransitionModel and DensityTransitionModel.

Tests: Not run locally.


FlorianPfaff:
Closing as old/low-value relative to the current queue: after updating with main it still has broad backend failures across NumPy/PyTorch/JAX, so it is not worth repairing in this maintenance pass.

PR #3199, opened 2026-06-26
Validate time-offset aggregate metric names

FlorianPfaff:
## Summary
- validate and normalize the exported `aggregate_time_offset_sweeps` metric argument before aggregation
- add regression coverage for invalid metric names and whitespace/case normalization
- include `std` in the aggregate validation fixture so the normalized metric path is exercised

## Testing
- `PYTHONPATH=/mnt/data/patch_check/src python -m pytest -q /mnt/data/patch_check/tests/calibration/test_time_offset_aggregate_validation.py`
- `python -m py_compile /mnt/data/patch_check/src/pyrecest/calibration/__init__.py`

Note: full repository pytest was not run locally because the sandbox cannot resolve github.com to clone/install the full repository dependency set.

### DIFF
--- a/src/pyrecest/calibration/__init__.py
+++ b/src/pyrecest/calibration/__init__.py
@@ -9,7 +9,7 @@ from .bias import (
 )
 from .time_offset import (
     TimeOffsetFitResult,
-    aggregate_time_offset_sweeps,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded: current main already contains the time-offset aggregate metric validation and normalization path, with broader preservation of summary metrics.

PR #3201, opened 2026-06-26
Improve NumPy linalg input handling

FlorianPfaff:


FlorianPfaff:
Closing as low-value/superseded: current main already coerces NumPy solve inputs with np.asarray, so this stale conflict branch no longer carries a distinct production fix.

PR #3212, opened 2026-06-27
Validate Autograd multinomial pvals

FlorianPfaff:
## Summary
- stop re-exporting the Autograd backend's raw multinomial sampler directly
- validate `pvals` before dispatch so non-real probability vectors fail with a clear `TypeError`
- add focused Autograd backend regression coverage

## Bug fixed
The Autograd backend still exposed `autograd.numpy.random.multinomial` without the probability validation added for the NumPy backend. Inputs with non-real probability dtypes could therefore be passed through to the low-level sampler instead of being rejected consistently by PyRecEst's backend contract.

## Validation
- Syntax-checked the modified source and new test file with `ast.parse` locally.
- Full pytest was not run locally because the sandbox cannot resolve `github.com` to clone/install the repository dependency set.

### DIFF
--- a/src/pyrecest/_backend/autograd/random.py
+++ b/src/pyrecest/_backend/autograd/random.py
@@ -1,6 +1,22 @@
 """Autograd based random backend."""
 
 import autograd.numpy as _np
-from autograd.numpy.random import get_state, multinomial, randint, seed, set_state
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded by the newer active multinomial probability-validation work in #3220, which covers JAX and PyTorch as well.

PR #3215, opened 2026-06-27
Fix padded history dtype validation

FlorianPfaff:
## Summary
- use dtype-kind validation for padded history arrays
- keep backend dtype fallback validation
- add a regression test for unsigned numeric history input

## Testing
- syntax checked the edited source and test snippets locally

Full pytest was not run locally because the sandbox cannot install the repository dependencies.


FlorianPfaff:
Closing as low-value for this maintenance pass: after main moved forward this branch still has backend failures, and the queue has newer active validation PRs with automerge enabled.

PR #3241, opened 2026-06-27
Reject boolean JAX normal scales

FlorianPfaff:
## Summary
- add a JAX `random.normal` scale validator that rejects boolean-valued scales before converting to JAX arrays
- preserve the existing negative-scale validation and NumPy-compatible sample-shape handling
- align the JAX backend with the existing NumPy/PyTorch normal-scale validation behavior

## Validation
- Compared `fix-jax-normal-boolean-scale` against `main`: one file changed, 15 additions / 3 deletions in `src/pyrecest/_backend/jax/random.py`.
- Locally imported the patched JAX random backend and verified boolean scales such as `True`, `False`, `[1.0, False]`, and JAX boolean arrays now raise `TypeError`.
- Verified `scale=-1.0` still raises `ValueError` and numeric array scales with `size=(3, 2)` still produce samples of shape `(3, 2)`.

Note: I attempted to add a regression test, but the GitHub connector safety layer blocked test-file writes for the executable test payload. The runtime fix itself is committed on this branch.

### DIFF
--- a/src/pyrecest/_backend/jax/random.py
+++ b/src/pyrecest/_backend/jax/random.py
@@ -304,11 +304,23 @@ def randint(low=None, high=None, size=None, *args, **kwargs):
     return set_state_return(has_state, state, res)
 
 
+def _validate_normal_scale(scale):
+    if _contains_boolean_value(scale):
+        raise TypeError("scale must be real numeric, not boolean")
+    try:
+        scale_array = _jnp.asarray(scale)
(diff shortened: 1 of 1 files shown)


FlorianPfaff:
Closing this as superseded by the newer boolean JAX normal-scale changes already on the merge queue/mainline; this branch is now conflicted and no longer worth repairing separately.

PR #3257, opened 2026-06-28
Validate DVS contour sample counts

FlorianPfaff:
## Summary
- validate DVS point-process contour sample counts as integer-like scalars greater than 2
- normalize accepted integer-like values to Python `int`
- add regression coverage for fractional/text contour sample count inputs

## Bug fixed
Previously, SCGP/DVS contour sample counts could be coerced via `int(...)`, so non-integer values such as `3.5` could silently become `3`. That changes the likelihood discretization without surfacing an invalid configuration.

## Testing
- Added focused unittest coverage in `tests/experimental/test_dvs_event_likelihood_counts.py`
- Targeted validation checked by inspection; full CI should run on the PR


FlorianPfaff:
Closing this as a no-op: the PR currently has no changed files after the branch/base updates, so there is nothing left to merge.

PR #3321, opened 2026-06-28
Support complex Hermitian covariance numerics

FlorianPfaff:
## Summary
- Allow native complex covariance-like matrices in `pyrecest.numerics` instead of rejecting complex dtypes up front.
- Use conjugate-transpose/Hermitian semantics for symmetry checks, symmetrization, PSD projection, and jittered Cholesky preprocessing.
- Add regression coverage for complex Hermitian covariance helpers while preserving rejection of unsupported object-valued inputs.

## Testing
- `PYTHONPATH=/tmp/pyrecest_test pytest -q /tmp/test_numerics.py` with a temporary package stub for local validation: 21 passed
- `python -m py_compile /tmp/numerics.py /tmp/test_numerics.py`

Note: full repository tests were not run in the connector environment.

### DIFF
--- a/src/pyrecest/numerics.py
+++ b/src/pyrecest/numerics.py
@@ -11,7 +11,7 @@ from pyrecest.exceptions import (
     ShapeError,
 )
 
-_UNSUPPORTED_NUMERIC_KINDS = {"b", "S", "U", "c", "M", "m"}
+_UNSUPPORTED_NUMERIC_KINDS = {"b", "S", "U", "M", "m"}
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this rather than continuing repairs: the complex Hermitian covariance numerics change is already failing multiple NumPy jobs and still has long backend jobs running, so it is not a good high-value repair target compared with the narrower PRs now in the automerge queue.

PR #3378, opened 2026-06-29
Fix JAX backend helper

FlorianPfaff:


FlorianPfaff:
Summary:
- Aligns the low-level JAX backend helper with the public facade when callers provide an optional destination array.
- Extends backend-portable regression coverage to exercise both public and low-level JAX calls.

Testing:
- Added focused coverage.
- Full local test execution was not available in this connector workflow; CI should run on the PR branch.

PR #3398, opened 2026-06-29
Fix PyTorch repeat axis handling

FlorianPfaff:
Fixes PyTorch backend repeat so NumPy-style axis arguments work consistently with the backend facade. Adds a focused regression test for scalar repeats, per-element repeat counts, flattening behavior, and raw backend access.

Testing: added a focused regression test; full test suite was not run in this connector environment.


FlorianPfaff:
Superseded by #3399, which carries the same fix on a fresh branch from the current main and is mergeable.

PR #3410, opened 2026-06-29
Fix raw PyTorch trace signature

FlorianPfaff:
## Summary
- Restore the raw PyTorch backend trace wrapper so `pyrecest._backend.pytorch.trace` accepts NumPy-style `offset`, `axis1`, `axis2`, `dtype`, and `out` arguments.
- Keep public `backend.trace` aligned when the active backend is PyTorch.
- Add focused regression coverage for the raw PyTorch entry point.

## Bug fixed
`pyrecest._backend.pytorch.trace` currently exposes the raw PyTorch-style helper, so calls with NumPy-compatible arguments such as `offset=`, `axis1=`, `axis2=`, `dtype=`, or `out=` fail even though PyRecEst's backend facade documents and tests a NumPy-style backend contract.

## Testing
- Restored `tests/backend_support/test_pytorch_trace_contract.py` from the validated raw-trace fix branch.
- The earlier branch passed MegaLinter; full CI should run on this PR.

### DIFF
--- a/src/pyrecest/__init__.py
+++ b/src/pyrecest/__init__.py
@@ -327,6 +327,46 @@ def _patch_pytorch_stack_helpers_facade() -> None:
         setattr(backend, helper_name, helper)
 
 
+def _patch_raw_pytorch_trace_facade() -> None:
+    """Make raw PyTorch ``trace`` follow NumPy's trace signature."""
+
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #3412, which contains a smaller current-main fix without merge conflicts. This raw-trace branch diverged from main and is not the active fix path.

PR #3436, opened 2026-06-29
Fix PyTorch cov integer promotion

FlorianPfaff:
## Summary
- Patch the raw PyTorch backend `cov` wrapper so integer observations are promoted to the PyTorch default floating dtype before calling `torch.cov`.
- Preserve existing floating/complex behavior and continue to patch the public facade when `PYRECEST_BACKEND=pytorch`.
- Add backend-portable regression coverage for both raw PyTorch use under a NumPy public backend and the public PyTorch backend.

## Rationale
`torch.cov` rejects integer inputs, while NumPy-style covariance accepts integer observations by promoting them for floating-point covariance computation. The previous wrapper passed integer tensors through directly, so `raw_pytorch.cov(raw_pytorch.array(..., dtype=int64), bias=True)` failed.

## Tests
- Not run locally in this environment; added `tests/backend_support/test_pytorch_cov_contract.py` for CI.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -310,6 +310,55 @@ def _patch_pytorch_clip_numpy_contract() -> None:
         backend.clip = clip
 
 
+def _patch_pytorch_cov_numpy_contract() -> None:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3437, opened 2026-06-29
Fix raw JAX matmul out under NumPy backend

FlorianPfaff:
## Summary
- Patch raw `pyrecest._backend.jax.matmul` from backend-support initialization so it honors `out` even when the active public backend is not JAX.
- Preserve the active JAX public facade update when JAX is selected.
- Add subprocess regression coverage for importing the raw JAX backend under the NumPy public backend.

## Bug fixed
A previous JAX `matmul(..., out=...)` fix only covered the active JAX public backend path. When `PYRECEST_BACKEND=numpy` and user code imports `pyrecest._backend.jax` directly, raw `matmul` still ignored `out`, so incompatible output shapes could silently succeed by returning the computed result. The new patch makes raw JAX `matmul` write through `.at[...].set(...)`, which also raises on incompatible `out` shapes.

## Testing
- Syntax-checked the changed source and added test content locally.
- Could not run the repository test suite locally because this container cannot resolve `github.com`; CI should execute the regression test.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -310,6 +310,32 @@ def _patch_pytorch_clip_numpy_contract() -> None:
         backend.clip = clip
 
 
+def _patch_raw_jax_matmul_out_contract() -> None:
+    """Make raw JAX matmul honor ``out`` under any active public backend."""
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3438, opened 2026-06-29
Fix PyTorch cov NumPy-style contract

FlorianPfaff:
## Summary
- Patch the PyTorch backend `cov` helper during backend-support initialization so it accepts NumPy-style `y`, `rowvar`, `ddof`, and `dtype` arguments.
- Preserve existing PyTorch-style `correction` support and old positional correction compatibility.
- Apply the shim to both the active public PyTorch backend and raw `pyrecest._backend.pytorch` after normal `pyrecest` import.
- Add focused subprocess regression coverage for public and raw PyTorch `cov` entry points.

## Bug fixed
The PyTorch backend exposed `cov(input, correction=..., fweights=..., aweights=..., bias=...)` as a thin `torch.cov` wrapper. Calls that are valid under the NumPy/PyRecEst contract, such as `backend.cov(x, y=y, rowvar=False, bias=True, dtype=backend.float64)` or `backend.cov(x, ddof=0)`, therefore failed or could not be expressed, while the JAX backend already accepts the broader NumPy-style signature.

## Testing
- Added `tests/backend_support/test_pytorch_cov_numpy_contract.py`.
- Not run locally in this connector session; CI should execute the focused regression.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -310,6 +310,146 @@ def _patch_pytorch_clip_numpy_contract() -> None:
         backend.clip = clip
 
 
+def _patch_pytorch_cov_numpy_contract() -> None:
+    """Make PyTorch cov accept NumPy-style y/rowvar/ddof/dtype arguments."""
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3439, opened 2026-06-29
Fix raw PyTorch diag under non-PyTorch backends

FlorianPfaff:
## Summary
- Patch raw `pyrecest._backend.pytorch.diag` from `pyrecest.backend_support` so it is fixed even when the active public backend is not PyTorch.
- Preserve the public PyTorch facade update when `PYRECEST_BACKEND=pytorch` is active.
- Add a regression covering `PYRECEST_BACKEND=numpy` followed by direct raw PyTorch `diag` usage with NumPy-style list inputs and `k=`.

## Bug fixed
`backend_tools._patch_pytorch_diag_numpy_contract()` only ran when the selected public backend was PyTorch. With `PYRECEST_BACKEND=numpy`, importing `pyrecest` and then using `pyrecest._backend.pytorch.diag([1, 2, 3], k=1)` still exposed raw `torch.diag`, which rejects the NumPy-style `k=` keyword/list-input contract.

## Testing
- Added `tests/backend_support/test_raw_pytorch_diag_default_backend.py`.
- Syntax-checked the modified Python content before committing; full test suite not run in this environment.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -310,6 +310,36 @@ def _patch_pytorch_clip_numpy_contract() -> None:
         backend.clip = clip
 
 
+def _patch_pytorch_diag_numpy_contract() -> None:
+    """Make PyTorch diag accept array-like inputs and NumPy's ``k`` keyword."""
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3444, opened 2026-06-29
Fix raw JAX std out contract

FlorianPfaff:
## Summary
- patch raw `pyrecest._backend.jax.std` from backend-support initialization so `out=` is handled even when the public backend is not JAX
- keep the public JAX facade aligned when JAX is the active backend
- add a subprocess regression for `PYRECEST_BACKEND=numpy` followed by direct raw JAX `std(..., out=...)`

## Bug fixed
The existing JAX `std(..., out=...)` compatibility patch only ran through the active public JAX facade path. With `PYRECEST_BACKEND=numpy`, importing `pyrecest` and then using `pyrecest._backend.jax.std(values, out=...)` could still expose `jax.numpy.std`, which rejects non-`None` `out` values. The raw backend now computes with `out=None` and returns the JAX `.at[...]` updated output array.

## Testing
- Not run locally in this connector session; added focused regression coverage in `tests/backend_support/test_raw_jax_std_out_default_backend.py`.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -340,6 +340,54 @@ def _patch_jax_outer_numpy_contract() -> None:
         backend.outer = outer
 
 
+def _patch_jax_std_out_numpy_contract() -> None:
+    """Make raw JAX std honor NumPy's ``out`` contract."""
+
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3448, opened 2026-06-29
Fix PyTorch concatenate axis=None contract

FlorianPfaff:
## Summary
- patch the PyTorch backend compatibility hook so `concatenate(..., axis=None)` flattens each input before concatenating, matching NumPy/JAX behavior
- keep the raw `pyrecest._backend.pytorch.concatenate` contract patched even when another public backend is selected
- add subprocess regression tests for both public PyTorch and raw PyTorch under a NumPy public backend

## Bug
On current `main`, `pyrecest._backend.pytorch.concatenate(..., axis=None)` forwards `dim=None` to `torch.cat`, which raises instead of applying NumPy's documented flattening behavior.

## Tests
- Not run locally in this environment; added targeted backend-portable regression tests for CI.


FlorianPfaff:

PR #3449, opened 2026-06-29
Fix PyTorch stack helpers with NumPy backend

FlorianPfaff:
## Summary
- Patch the low-level PyTorch stack helper entry points from the backend support compatibility hook.
- Preserve NumPy-style array-like sequence support even when the selected public backend is NumPy.
- Add a subprocess regression for `hstack`, `vstack`, `column_stack`, and `dstack` with `PYRECEST_BACKEND=numpy`.

## Bug fixed
When the process selected the NumPy public backend, the PyTorch-specific stack-helper normalization path did not run. Direct PyTorch backend helper calls could therefore still reach Torch-native stack functions, which reject Python list elements such as `([1, 2], [3, 4])`.

## Testing
- Added `tests/backend_support/test_pytorch_stack_helpers_raw_backend_contract.py`.
- Not run locally in this connector-only session; CI should execute the focused backend-portable regression.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -1,6 +1,53 @@
-"""PyTorch dtype promotion compatibility helpers."""
+"""PyTorch backend compatibility helpers."""
 
 from __future__ import annotations
+
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3455, opened 2026-06-29
Fix JAX matmul backend contract

FlorianPfaff:


FlorianPfaff:
Closing this conflicted backend-contract PR as old/low-value for this maintenance sweep, per the PyRecEst cleanup guidance.

PR #3456, opened 2026-06-29
Fix PyTorch arctan array-like backend contract

FlorianPfaff:
## Summary
- Patch the PyTorch backend-support hook so `arctan` accepts NumPy-style array-like inputs instead of exposing bare `torch.arctan`.
- Keep the raw PyTorch backend and active public PyTorch facade synchronized.
- Add focused subprocess regression coverage for both public PyTorch and raw PyTorch under a NumPy public backend.

## Bug fixed
`pyrecest._backend.pytorch.arctan` was still a direct `torch.arctan` export. Calls such as `raw_pytorch.arctan([0.0, 1.0])` failed with `TypeError`, while the NumPy/JAX-style backend contract accepts array-like inputs.

## Validation
- `python -m py_compile /tmp/_torch_dtype_promotion_contract.py /tmp/test_pytorch_arctan_contract.py`
- Local smoke check confirmed bare `torch.arctan([0.0, 1.0])` rejects lists and the patched wrapper accepts them.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -1,15 +1,10 @@
-"""PyTorch dtype promotion compatibility helpers."""
+"""PyTorch backend compatibility helpers."""
 
 from __future__ import annotations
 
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3457, opened 2026-06-29
Patch raw PyTorch repeat across public backends

FlorianPfaff:
## Summary
- patch raw `pyrecest._backend.pytorch.repeat` from the backend-support compatibility path, so the raw PyTorch helper gets NumPy-style repeat semantics even when the selected public backend is NumPy
- keep the public NumPy backend untouched while still aligning `backend.repeat` when the public backend is PyTorch
- add a focused subprocess regression for importing raw PyTorch repeat under `PYRECEST_BACKEND=numpy`

## Bug fixed
`pyrecest._backend.pytorch.repeat` was previously only adapted by `_adapt_pytorch_repeat_contract()` when `PYRECEST_BACKEND=pytorch`. With the default/NumPy public backend, direct raw PyTorch backend users could still reach bare `torch.repeat_interleave`, which rejects list inputs / NumPy-style repeat counts and does not accept the `axis=` alias used by PyRecEst's backend contract.

## Testing
- Added `tests/backend_support/test_raw_pytorch_repeat_default_backend.py`
- Verified the committed branch files via GitHub fetch after writing them
- Full test suite was not run in this connector environment because the execution container cannot resolve `github.com` for cloning/installing the repository

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -1,15 +1,111 @@
-"""PyTorch dtype promotion compatibility helpers."""
+"""PyTorch backend compatibility helpers."""
 
 from __future__ import annotations
 
+from operator import index as _operator_index
+
+
+def _pytorch_repeat_count(repetition) -> int:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3459, opened 2026-06-29
Fix PyTorch argsort NumPy-style contract

FlorianPfaff:
## Summary
- Patch PyTorch `argsort` during backend initialization so it accepts array-like inputs and NumPy-style `axis`, including `axis=None` flattening.
- Preserve PyTorch-style `dim` and `descending` arguments while mapping stability-related NumPy options onto `torch.argsort`.
- Add subprocess regression tests for both the active public PyTorch backend and raw PyTorch backend when the public backend is NumPy.

## Bug
`pyrecest._backend.pytorch.argsort` was exposed directly from `torch.argsort`, so calls that are valid for the NumPy backend, for example `argsort([[3, 1, 2]], axis=1)` or `axis=None`, failed instead of following the backend facade contract.

## Validation
- Ran an isolated local check of the new wrapper logic against installed PyTorch.
- Full repository tests were not run locally because the GitHub repository could not be cloned from this environment.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -1,15 +1,12 @@
-"""PyTorch dtype promotion compatibility helpers."""
+"""PyTorch backend compatibility helpers."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3460, opened 2026-06-29
Patch raw PyTorch round under NumPy backend

FlorianPfaff:
## Summary
- Move the PyTorch `round` NumPy-compatibility wrapper into the backend-support import hook so the raw PyTorch backend is patched even when the selected public backend is NumPy.
- Preserve active public PyTorch facade behavior when `PYRECEST_BACKEND=pytorch`.
- Add regression coverage for importing PyRecEst with `PYRECEST_BACKEND=numpy` and then calling `pyrecest._backend.pytorch.round` with list inputs, `decimals`, and `out`.

## Bug fixed
The earlier PyTorch `round` compatibility patch only covered the active PyTorch backend path. With `PYRECEST_BACKEND=numpy`, importing `pyrecest` and then using `pyrecest._backend.pytorch.round([1.24, 2.76], decimals=1)` still exposed raw `torch.round`, which rejects NumPy-style array-like inputs.

## Testing
- Syntax-checked the modified backend-support module and the new regression test locally.
- Full test suite not run in this environment; CI should validate the complete matrix.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -83,7 +83,6 @@ def _patch_pytorch_dot_numpy_contract() -> None:
         ModuleNotFoundError
     ):  # pragma: no cover - PyTorch backend import failed earlier
         return
-
     original_dot = raw_pytorch.dot
     if getattr(original_dot, "_pyrecest_numpy_contract", False):
         return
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

FlorianPfaff:
Closing this old conflicted optional PyRecEst backend-compatibility PR as low-value for now.

PR #3461, opened 2026-06-29
Fix PyTorch like helper array-like contract

FlorianPfaff:
Patch PyTorch backend like helpers to accept array-like inputs and add focused regression tests.


FlorianPfaff:

PR #3462, opened 2026-06-29
Fix PyTorch special arraylike contract

FlorianPfaff:
Patch PyTorch special helper wrappers so arraylike inputs work consistently. Adds a focused regression test for the default NumPy public backend path. Local validation: syntax checks and a PyTorch smoke check for list inputs.


FlorianPfaff:

PR #3464, opened 2026-06-29
Fix public PyTorch dtype promotion facade

FlorianPfaff:
## Summary
- Synchronize the active public PyTorch facade when the dtype-promotion compatibility patch replaces raw `pyrecest._backend.pytorch.convert_to_wider_dtype`.
- Preserve the early-return/idempotent path by also syncing `pyrecest.backend.convert_to_wider_dtype` when the raw PyTorch backend has already been patched.
- Add a focused regression covering the reused-patch synchronization path.

## Bug fixed
`patch_pytorch_dtype_promotion_contract()` updated the raw PyTorch backend implementation but left an already-created `pyrecest.backend.convert_to_wider_dtype` facade attribute stale when `PYRECEST_BACKEND=pytorch`. Public PyTorch users could therefore still hit the old dtype-order table instead of Torch's promotion rules.

## Testing
- Syntax-compiled the modified helper and new regression test locally with `py_compile`.
- Full repository test suite was not run locally because this environment could not clone GitHub repositories via DNS.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 def patch_pytorch_dtype_promotion_contract() -> None:
     """Make PyTorch mixed-dtype helpers use Torch's promotion rules."""
     try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3466, opened 2026-06-29
Fix PyTorch array_equal equal_nan contract

FlorianPfaff:
## Summary
- Add a PyTorch compatibility shim for `array_equal(..., equal_nan=True)`.
- Patch both the raw PyTorch backend and the active public PyTorch facade.
- Add regression tests for the raw backend under NumPy selection and for the public PyTorch facade.

## Validation
- Python syntax check for the changed files.
- Local Torch smoke check for the new equality semantics.


FlorianPfaff:

PR #3468, opened 2026-06-29
Fix PyTorch diff NumPy-style contract

FlorianPfaff:
## Summary
- patch raw `pyrecest._backend.pytorch.diff` so it accepts array-like inputs and NumPy-style `axis`, while preserving PyTorch `dim`
- support scalar `prepend`/`append` broadcasting and `out=` handling
- keep the public PyTorch facade synchronized and add regressions for public PyTorch plus raw PyTorch under a NumPy public backend

## Bug fixed
`pyrecest._backend.pytorch.diff` was still exposed as raw `torch.diff`. That rejects Python-list inputs and does not accept NumPy-style `axis`, so backend-portable code that works with the NumPy facade could fail when switching to PyTorch or when using the raw PyTorch backend after normal PyRecEst import.

## Validation
- `python -m py_compile /tmp/backend_tools_new.py /tmp/test_pytorch_diff_contract.py`
- local PyTorch smoke check for list input, `axis=`, scalar `prepend`, `dim=` alias, and `append=`

Full repository tests were not run in this connector environment; CI should run the focused regression.

### DIFF
--- a/src/pyrecest/backend_tools.py
+++ b/src/pyrecest/backend_tools.py
@@ -280,6 +280,83 @@ def _patch_pytorch_stack_helpers_numpy_contract() -> None:
         setattr(pytorch_backend, helper_name, helper)
 
 
+def _patch_pytorch_diff_numpy_contract() -> None:
+    """Make PyTorch diff accept NumPy-style array-like inputs and ``axis``."""
+
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3472, opened 2026-06-29
Restore PyTorch allclose equal_nan contract

FlorianPfaff:
## Summary
- restore the PyTorch backend `allclose(..., equal_nan=...)` compatibility contract
- patch both raw `pyrecest._backend.pytorch.allclose` and the active public PyTorch facade through the backend-support hook
- add a focused backend-support regression for NaN handling and broadcasted array-like inputs

## Bug fixed
Current `main` still exposes the PyTorch backend `allclose` wrapper with only `(a, b, atol, rtol)`. Backend-portable code using the NumPy/PyTorch missing-value comparison keyword, for example `backend.allclose([nan], [nan], equal_nan=True)`, therefore raises under PyTorch instead of matching the expected contract.

## Validation
- `python -m py_compile /tmp/_torch_dtype_promotion_contract_new.py /tmp/test_pytorch_allclose_contract.py`
- local Torch smoke check of the wrapper logic for `equal_nan=True`, `equal_nan=False`, and broadcasted array-like inputs
- verified changed files by fetching the branch contents from GitHub

Full repository tests were not run in this connector environment; CI should run the focused regression.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -14,11 +14,10 @@ def _patch_pytorch_allclose_equal_nan_contract(raw_pytorch, torch_module) -> Non
     default_rtol = raw_pytorch.rtol
 
     def _as_allclose_tensor(value, *, device):
-        if torch_module.is_tensor(value):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3473, opened 2026-06-29
Fix PyTorch conj array-like backend contract

FlorianPfaff:
## Summary
- Patch PyTorch `conj` compatibility handling so both raw `pyrecest._backend.pytorch.conj` and the public PyTorch backend accept NumPy-style array-like inputs.
- Preserve `out=` handling for both torch tensors and NumPy arrays.
- Add focused subprocess regressions for raw PyTorch under a NumPy public backend and for the public PyTorch backend.

## Bug fixed
`pyrecest._backend.pytorch.conj` still exposed bare `torch.conj`, so calls such as `raw_pytorch.conj([1.0 + 2.0j])` failed even though PyRecEst's backend facade generally accepts array-like inputs. The public PyTorch facade had the same issue for `backend.conj([...])`.

## Validation
- `python -m py_compile /tmp/backend_tools_new.py`
- Local standalone Torch smoke check confirmed bare `torch.conj` rejects Python lists, while the patched wrapper accepts list inputs and both torch/NumPy `out=` destinations.
- Full repository test suite was not run locally because this sandbox cannot clone GitHub via DNS; CI should run on GitHub.

### DIFF
--- a/src/pyrecest/backend_tools.py
+++ b/src/pyrecest/backend_tools.py
@@ -326,6 +326,39 @@ def _patch_raw_pytorch_comparison_numpy_contract() -> None:
             setattr(backend, helper_name, helper)
 
 
+def _patch_raw_pytorch_conj_numpy_contract() -> None:
+    """Make raw PyTorch conj accept NumPy-style array-like inputs."""
+
+    try:
+        import pyrecest._backend.pytorch as pytorch_backend  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value/conflicting after the backend-contract cleanup train; this older PyTorch conj micro-fix conflicts and still carries failing checks, while related backend fixes are already being merged through current main.

PR #3474, opened 2026-06-29
Fix PyTorch isclose equal_nan contract

FlorianPfaff:
## Summary
- Patch the PyTorch backend `isclose` helper to accept NumPy/PyTorch's `equal_nan` keyword.
- Preserve array-like coercion, dtype promotion, and active public-backend replacement when PyTorch is selected.
- Add subprocess regression tests for both public PyTorch backend use and raw PyTorch backend use while NumPy is the active public backend.

## Bug
`pyrecest._backend.pytorch.isclose` exposed a wrapper without `equal_nan`, so NaN-aware comparisons raised `TypeError` instead of matching the NumPy/JAX backend contract.

## Tests
- Added `tests/backend_support/test_pytorch_isclose_contract.py`
- Not run locally in this environment; CI should exercise the added regression tests.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -3,6 +3,52 @@
 from __future__ import annotations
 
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3475, opened 2026-06-29
Fix PyTorch predicate array-like backend contract

FlorianPfaff:
## Summary
- Patch raw PyTorch predicate helpers `isfinite`, `isinf`, `isnan`, and `isreal` so they accept NumPy-style array-like inputs.
- Preserve active public PyTorch backend parity by syncing the patched helpers onto `pyrecest.backend` when `PYRECEST_BACKEND=pytorch`.
- Add regression coverage for public PyTorch and raw PyTorch usage under the NumPy public backend, including `out=` handling for Torch and NumPy outputs.

## Bug fixed
The PyTorch backend still exposed bare Torch predicate helpers. Calls that are valid for NumPy-style backend-portable code, such as `pyrecest._backend.pytorch.isfinite([0.0, float("inf")])` or `backend.isnan([...], out=...)`, could fail under PyTorch because Torch requires tensor input and these helpers do not accept NumPy's `out=` keyword.

## Validation
- `python -m py_compile /tmp/_torch_dtype_promotion_contract.py /tmp/test_pytorch_predicate_contract.py`
- Standalone local Torch smoke test for list inputs plus Torch/NumPy `out=` destinations.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -3,34 +3,79 @@
 from __future__ import annotations
 
 
+def _return_or_store_predicate_out(raw_pytorch, result, out):
+    """Return ``result`` or store it through a NumPy/PyTorch-style ``out``."""
+
+    if out is None:
+        return result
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3482, opened 2026-06-29
Fix PyTorch log1p array-like backend contract

FlorianPfaff:
## Summary
- Add an import-time PyTorch `log1p` adapter that coerces NumPy-style array-like inputs through the PyTorch backend before calling the raw helper.
- Preserve `out=` handling for tensor-like and mutable array outputs.
- Add regression coverage for raw PyTorch `log1p` with array-like inputs and `out=`.

## Bug fixed
`pyrecest._backend.pytorch.log1p` was still exposed as raw `torch.log1p`, unlike neighboring unary helpers such as `log`, `sqrt`, and `exp`. Direct raw-backend calls such as `pytorch_backend.log1p([0.0, 1.0, 3.0])` could therefore fail because Torch expects a tensor rather than a Python list. The wrapper keeps the raw backend and the public PyTorch facade aligned when PyTorch is selected.

## Testing
- Added `test_raw_pytorch_log1p_accepts_array_like_inputs_and_out` in `tests/backend_support/test_pytorch_linear_helpers_contract.py`.
- Not run locally in this connector session; CI should run the focused backend-portable regression.

### DIFF
--- a/src/pyrecest/_backend_submodules.py
+++ b/src/pyrecest/_backend_submodules.py
@@ -97,6 +97,35 @@ def _adapt_pytorch_allclose_keyword_contract(backend: ModuleType) -> None:
     setattr(pytorch_backend, "allclose", wrapped_allclose)
 
 
+def _adapt_pytorch_log1p_contract(backend: ModuleType) -> None:
+    """Adapt PyTorch ``log1p`` to accept NumPy-style array-like inputs."""
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this conflicted, low-value maintenance PR because automerge cannot proceed without substantial conflict repair.

PR #3492, opened 2026-06-30
Restore raw PyTorch repeat default-backend contract

FlorianPfaff:
## Summary
- patch raw `pyrecest._backend.pytorch.repeat` from the backend-support compatibility path
- keep NumPy-style array-like inputs, repeat counts, and `axis`/`dim` handling available even when the public backend is NumPy
- add a subprocess regression for direct raw PyTorch `repeat` under `PYRECEST_BACKEND=numpy`

## Bug fixed
`pyrecest._backend.pytorch.repeat` was only adapted by the backend-submodule compatibility layer when the active public backend was PyTorch. Under the default/NumPy public backend, direct raw PyTorch backend users could still reach bare `torch.repeat_interleave`, which rejects list inputs / NumPy-style repeat counts and does not accept PyRecEst's `axis=` alias.

## Testing
- Not run locally in this connector session; added `tests/backend_support/test_raw_pytorch_repeat_default_backend.py` for the failing import/backend combination.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -1,15 +1,111 @@
-"""PyTorch dtype promotion compatibility helpers."""
+"""PyTorch backend compatibility helpers."""
 
 from __future__ import annotations
 
+from operator import index as _operator_index
+
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3493, opened 2026-06-30
Fix raw JAX matmul out contract

FlorianPfaff:
## Summary
- patch the raw JAX backend `matmul` wrapper so `out` is honored even when JAX is not the selected public backend
- keep the selected public JAX facade in sync when it is active
- add direct-backend regression coverage for `out` dtype/shape behavior

## Bug fixed
The current JAX backend implementation computes `result = jnp.matmul(x, y)` and returns `result` even when `out` is supplied. That means invalid `out` shapes can be silently ignored instead of following the backend contract used by the existing `trace`, `argmax`, and `argmin` wrappers, which write through `.at[...].set(...)` and raise on incompatible shapes.

## Validation
- Verified locally in an isolated JAX snippet that `out.at[...].set(result)` returns the expected values and raises `ValueError` for incompatible shapes.
- Existing focused regression coverage is included in this PR.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -73,7 +73,6 @@ def _patch_pytorch_dot_numpy_contract() -> None:
         import pyrecest.backend as backend  # pylint: disable=import-outside-toplevel
     except ModuleNotFoundError:  # pragma: no cover - import fails before this module
         return
-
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

FlorianPfaff:
Closing this conflicted, low-value maintenance PR because automerge cannot proceed without substantial conflict repair.

PR #3503, opened 2026-06-30
Fix PyTorch unary predicate array-like contract

FlorianPfaff:
## Summary
- Patch raw PyTorch `isfinite`, `isinf`, `isnan`, and `isreal` so they accept PyRecEst/NumPy-style array-like inputs.
- Keep the active public PyTorch facade aligned while leaving the NumPy public backend untouched.
- Add subprocess regressions for raw PyTorch under the NumPy public backend and for the public PyTorch backend.

## Bug fixed
The PyTorch backend exposed these unary predicate helpers directly from `torch`. Calls such as `pyrecest._backend.pytorch.isfinite([1.0, float("nan")])` therefore raised `TypeError`, even though PyRecEst backend helpers generally accept array-like inputs.

## Testing
- Local sanity check: bare Torch rejects list inputs for `isfinite`, `isinf`, `isnan`, and `isreal`.
- Local syntax check: `python -m py_compile` on the modified backend-support module and new test file.
- Full repository test suite not run in this connector session; CI should run the focused regressions.

### DIFF
--- a/src/pyrecest/backend_support/__init__.py
+++ b/src/pyrecest/backend_support/__init__.py
@@ -312,6 +312,46 @@ def _patch_pytorch_clip_numpy_contract() -> None:
         backend.clip = clip
 
 
+def _patch_pytorch_unary_predicate_numpy_contract() -> None:
+    """Make PyTorch unary predicates accept NumPy-style array-like inputs."""
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3506, opened 2026-06-30
Fix JAX take out contract

FlorianPfaff:
Summary:
- Wrap the JAX take helper so the result is computed first and then copied into the requested output object.
- Preserve the existing JAX take keyword handling.
- Add a subprocess regression for the public JAX backend.

Testing:
- Not run locally; regression added for CI.


FlorianPfaff:

PR #3511, opened 2026-06-30
Fix PyTorch assignment with NumPy boolean masks

FlorianPfaff:
## Summary
- Normalize NumPy boolean scalar/array indices before the PyTorch assignment helpers classify indices.
- Keep raw PyTorch assignment helpers and the active public PyTorch backend facade aligned after import.
- Add public and raw-backend regressions for NumPy boolean masks.

## Bug fixed
The PyTorch assignment helpers recognized Python/list and Torch boolean masks, but not NumPy boolean ndarrays. A call such as `backend.assignment(x, values, np.array([...], dtype=bool))` could be treated as integer-style indexing or rejected before mask assignment, while equivalent list-of-list masks already worked.

## Testing
- Not run locally; CI should run the new focused regressions.


FlorianPfaff:

PR #3513, opened 2026-06-30
Fix PyTorch comparison array-like/device contract

FlorianPfaff:
## Summary
- Patch raw PyTorch comparison helpers `greater`, `less`, `less_equal`, and `logical_or` to coerce NumPy-style array-like operands before calling Torch.
- Keep the public PyTorch backend facade aligned when `PYRECEST_BACKEND=pytorch` is active.
- Add subprocess regression coverage under both NumPy and PyTorch backend selections.

## Bug fixed
`less_equal` could leave array-like operands on CPU when the other operand was already a tensor on another device. The new wrapper normalizes operands onto the tensor operand's device before calling Torch.

## Validation
- Added comparison contract coverage to `tests/backend_support/test_pytorch_pad_contract.py`.
- Added a `meta`-device regression for `less_equal`.
- Full repository test suite not run locally in this environment.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -2,7 +2,18 @@
 
 from __future__ import annotations
 
-from operator import index as _operator_index
+from pyrecest.backend_support._torch_comparison_contract import (
+    patch_pytorch_comparison_arraylike_contract,
+)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this conflicted backend micro-PR as low-value to repair further after nearby backend changes merged; keeping the automerge-enabled queue for the remaining relevant changes.

PR #3514, opened 2026-06-30
Fix PyTorch array_from_sparse duplicate semantics

FlorianPfaff:
## Summary
- Patch the PyTorch backend compatibility layer so `array_from_sparse` reconstructs dense arrays by flat assignment instead of sparse COO densification.
- Preserve NumPy-style coordinate shape/bounds checks for sparse coordinates.
- Add raw and public PyTorch regressions for duplicate sparse indices.

## Bug fixed
The shared NumPy backend uses assignment-style reconstruction via `put(ravel_multi_index(...), data)`, so duplicate sparse coordinates follow last-write-wins assignment semantics. The PyTorch backend used `torch.sparse_coo_tensor(...).to_dense()`, which sums duplicate coordinates instead. Equivalent backend calls could therefore produce different dense arrays.

## Validation
- Syntax-checked the edited backend-support module and new regression file locally with `python -m py_compile`.
- Ran a focused local smoke check of the patched helper with duplicate vector and matrix coordinates.
- Full repository tests were not run locally in this connector session; CI should run the added regressions.


FlorianPfaff:
Closing this conflicted backend micro-PR as low-value to repair further after nearby backend changes merged; keeping the automerge-enabled queue for the remaining relevant changes.

PR #3522, opened 2026-06-30
Fix PyTorch take axis validation

FlorianPfaff:
## Summary
- Patch the PyTorch backend compatibility layer so `take(..., axis=...)` uses integer-index semantics instead of `int(axis)` coercion.
- Reject logical and non-integer axes such as `True`, `1.0`, `"1"`, `np.array(True)`, and `torch.tensor(True)` for both public and raw PyTorch `take`.
- Add subprocess regression coverage while preserving 0-D NumPy integer scalar axis support.

## Bug fixed
The raw PyTorch `take` helper normalized `axis` through `int(axis)`, so castable non-integer values could be silently accepted as dimensions. NumPy rejects those values, so a call such as `backend.take(values, [0], axis=1.0)` could incorrectly behave as `axis=1` under the PyTorch backend.

## Validation
- `python -m py_compile /mnt/data/_torch_dtype_promotion_contract.py`
- `python -m py_compile /mnt/data/test_pytorch_take_axis_contract.py`
- Full repository tests were not run locally because this execution container cannot resolve `github.com` to clone/install the repository; CI should run the added focused regression.


FlorianPfaff:

PR #3525, opened 2026-06-30
Fix PyTorch one_hot integer tensor labels

FlorianPfaff:
## Summary
- Patch the PyTorch backend compatibility layer so `one_hot` casts integer tensor label dtypes to `torch.long` before calling `torch.nn.functional.one_hot`.
- Keep the public `pyrecest.backend` facade and raw `pyrecest._backend.pytorch` helper in sync.
- Add subprocess regression coverage for `torch.int32`/`torch.int16` tensor labels while preserving `uint8` indicators.

## Bug fixed
The PyTorch backend accepted Python list labels because it converted them with `torch.LongTensor`, but it rejected integer tensors such as `torch.tensor([0, 2], dtype=torch.int32)`. This made valid integer-label inputs fail only when they were already tensors, even though the shared backend contract accepts integer label arrays.

## Testing
- `python -m py_compile /tmp/_torch_dtype_promotion_contract.py /tmp/test_pytorch_one_hot_contract.py`
- Local torch smoke check confirmed `torch.nn.functional.one_hot(torch.tensor([0], dtype=torch.int32), 3)` raises before casting and succeeds after casting to `torch.long`.

Full repository tests were not run locally in this connector-driven session.


FlorianPfaff:

PR #3526, opened 2026-06-30
Fix PyTorch scalar array reduction axes

FlorianPfaff:
## Summary
- Add a PyTorch reduction-axis compatibility helper that normalizes scalar NumPy-array and scalar Torch tensor axes before dispatching to Torch.
- Wire the helper into the existing PyTorch backend-support patch path.
- Add focused subprocess regressions for public PyTorch backend calls, raw PyTorch backend calls while the public backend is NumPy, and boolean scalar-axis rejection.

## Bug fixed
The PyTorch backend could forward scalar array-like axis values directly to Torch reductions. Calls such as `backend.sum(values, axis=np.array(1))` should follow NumPy-style scalar integer axis semantics, but direct Torch dispatch can fail for these inputs. The patch normalizes valid scalar integer axes to Python `int` and keeps boolean scalar axes invalid.

## Validation
- `python -m py_compile /mnt/data/_torch_dtype_promotion_contract_minimal.py /mnt/data/_torch_reduction_axis_contract.py /mnt/data/test_pytorch_reduction_axis_contract.py`
- Repository diff checked against `main` with the GitHub connector.
- Full repository test suite was not run locally because this environment could not clone GitHub (`Could not resolve host: github.com`); CI should run the added focused regressions.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract.py
@@ -4,6 +4,10 @@ from __future__ import annotations
 
 from operator import index as _operator_index
 
+from ._torch_reduction_axis_contract import (
+    patch_pytorch_reduction_axis_contract as _patch_pytorch_reduction_axis_contract,
+)
+
 
 def patch_pytorch_dtype_promotion_contract() -> None:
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3530, opened 2026-06-30
Reject unsupported unary transition sample counts

FlorianPfaff:
## Summary
- Validate transition sampler count requests for callbacks that only accept `state`.
- Allow the default count of one.
- Raise `ValueError` for larger requested counts instead of ignoring the request.
- Add regression tests for `SampleableTransitionModel` and `DensityTransitionModel`.

## Validation
- `python -m py_compile /tmp/_sampleable_transition_validation.py /tmp/test_transition_unary_sampler_count_contract.py`

### DIFF
--- a/src/pyrecest/models/_sampleable_transition_validation.py
+++ b/src/pyrecest/models/_sampleable_transition_validation.py
@@ -4,8 +4,15 @@ from __future__ import annotations
 
 from typing import Any
 
-from .likelihood import SampleableTransitionModel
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3555. The newer branch covers the same unsupported transition sample-count contract more generally for both sampleable and density transition models, and it has just been rebased onto current main for a fresh CI run.

PR #3546, opened 2026-06-30
Reject boolean PyTorch creation shapes

FlorianPfaff:
## Summary
- Add a PyTorch creation-helper compatibility layer that rejects boolean shape values before the base NumPy-style wrapper normalizes them.
- Cover `empty`, `zeros`, `ones`, and `full` for raw PyTorch backend imports and the public PyTorch backend.
- Keep valid integer shapes unchanged.

## Bug fixed
The PyTorch backend support wrapper normalized creation-helper shapes with `operator.index(...)`. Because Python booleans implement the integer index protocol, calls such as `backend.zeros(True)` or `raw_pytorch_backend.full([True], 1)` were accepted as shape `(1,)`. NumPy and raw PyTorch both reject boolean dimensions, so backend-portable code could silently allocate an unintended shape instead of surfacing a bad configuration.

## Validation
- Inspected the branch diff with `compare_commits`.
- Added focused backend-portable regression tests.
- Full suite not run locally because this connector environment cannot clone GitHub directly.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
@@ -33,6 +33,7 @@ def patch_pytorch_dtype_promotion_contract() -> None:
         return
 
     _patch_pytorch_logical_device_contract(raw_pytorch, backend, torch)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3579, opened 2026-07-01
Normalize NumPy array axes in PyTorch linalg norm

FlorianPfaff:
## Summary
- Normalize single-element NumPy array axes before passing them to `torch.linalg.norm`.
- Preserve PyTorch handling for non-array axes and multi-element array axes.
- Add a focused PyTorch backend regression for `axis=np.array(0)` and `axis=np.array([0])`.

## Bug fixed
`pyrecest._backend.pytorch.linalg.norm(values, axis=np.array(0))` currently forwards the NumPy array directly as `dim`, which triggers a PyTorch argument-parser/internal error instead of behaving like a NumPy-style scalar axis.

## Validation
- Verified the minimal failure/fix locally against installed PyTorch by normalizing `np.array(0)` and `np.array([0])` before `torch.linalg.norm`.
- Not run: full PyRecEst test suite; repository checkout was not available in this execution environment.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -1,5 +1,7 @@
 """Pytorch based linear algebra backend."""
 
+import operator as _operator
+
 import numpy as _np
 import scipy as _scipy
 import torch as _torch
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3581, opened 2026-07-01
Normalize JAX linalg norm array axes

FlorianPfaff:
## Summary
- Normalize NumPy scalar and one-element array `axis` values before delegating `pyrecest._backend.jax.linalg.norm` to JAX.
- Add a focused regression covering `axis=np.array(1)` and `axis=np.array([1])`.

## Bug fixed
`jax.numpy.linalg.norm` treats `axis` as a static argument and rejects non-hashable NumPy arrays before it can apply NumPy-compatible scalar-axis coercion. That made the PyRecEst JAX backend fail for NumPy-style scalar-array axes even though the NumPy contract accepts them.

## Validation
- Reproduced the raw JAX failure locally with `jnp.linalg.norm(..., axis=np.array(1))` and `axis=np.array([1])`.
- Verified the normalizer returns the expected `[5, 5]` result in an isolated JAX snippet.
- Not run: full pytest suite in this connector session.

### DIFF
--- a/src/pyrecest/_backend/jax/linalg.py
+++ b/src/pyrecest/_backend/jax/linalg.py
@@ -1,5 +1,6 @@
 """JAX-based linear algebra backend."""
 
+import numpy as _np
 import jax.numpy as _jnp
 import jax.scipy.linalg as _jax_scipy_linalg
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3586, opened 2026-07-01
Normalize PyTorch matrix_power scalar exponents

FlorianPfaff:
## Summary
- Normalize scalar NumPy array and scalar PyTorch tensor exponents before calling `torch.linalg.matrix_power`.
- Preserve the existing array-like matrix promotion behavior.
- Add a focused PyTorch backend regression for `np.array(2)` and `torch.tensor(2)` exponents.

## Bug fixed
`pyrecest.backend.linalg.matrix_power` already promotes array-like matrix inputs in the PyTorch backend, but it passed the exponent `n` through unchanged. Raw `torch.linalg.matrix_power` rejects scalar array/tensor exponent inputs even when they contain a valid integer value, so NumPy-style scalar array exponents failed despite being safely coercible to a Python integer.

## Validation
- Reproduced raw PyTorch failure locally for `torch.linalg.matrix_power(torch.eye(2), np.array(2))`.
- Verified locally that applying `operator.index(...)` after scalar extraction fixes NumPy scalar array and PyTorch scalar tensor exponents.
- Not run: full repository test suite in this connector session.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -130,9 +130,26 @@ def expm(a):
     return _torch.linalg.matrix_exp(_as_linalg_tensor(a))
 
 
+def _normalize_matrix_power_exponent(n):
+    """Return a Python integer exponent from scalar integer array inputs."""
+    if _torch.is_tensor(n):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3617, opened 2026-07-01
Fix PyTorch moveaxis NumPy-style axes

FlorianPfaff:
## Summary
- Patch the PyTorch backend support hook so `moveaxis` converts array-like inputs through the backend array helper before dispatching to `torch.moveaxis`.
- Normalize NumPy/PyTorch/list axis arguments to Python `int`/`tuple[int, ...]`, matching the NumPy-style backend contract.
- Add backend-portable regressions for the public PyTorch backend and for direct raw PyTorch backend use under the default backend.

## Bug fixed
`pyrecest._backend.pytorch.moveaxis` was still a direct `torch.moveaxis` export. Calls that are valid under NumPy/PyRecEst-style semantics, such as list inputs or NumPy scalar/vector axis arrays, failed with PyTorch argument-type errors.

## Validation
- `python -m py_compile /mnt/data/pytorch_dtype_contract_updated.py /mnt/data/test_pytorch_moveaxis_contract.py`
- Local isolated smoke check confirmed bare `torch.moveaxis` rejects list inputs and NumPy array axes before the compatibility wrapper.
- Full repository pytest was not run in this connector environment because the local container cannot clone/resolve `github.com`; CI should run the added focused regressions.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import importlib.util
 from pathlib import Path
+from operator import index as _operator_index
 
 
 def _load_base_contract_module():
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3626, opened 2026-07-01
Default missing summarize MTT flag

FlorianPfaff:
## Summary
- Default missing `scenario_config["mtt"]` to `False` in `summarize_filter_results()`.
- Add regression coverage for single-target Euclidean summaries with no explicit MTT flag.

## Bug fixed
Plain single-target scenario configs can include a manifold without an `mtt` key. `summarize_filter_results()` previously indexed `scenario_config["mtt"]` unconditionally before selecting the mean extractor, so those configs raised `KeyError` instead of being treated as non-MTT.

## Validation
- `python -m py_compile /tmp/summarize_filter_results.py /tmp/test_evaluation_summarize_missing_mtt.py`
- Full test suite not run locally because this execution container cannot resolve `github.com` for cloning/installing the repository; CI should run the focused regression.

### DIFF
--- a/src/pyrecest/evaluation/summarize_filter_results.py
+++ b/src/pyrecest/evaluation/summarize_filter_results.py
@@ -38,7 +38,7 @@ def summarize_filter_results(
         warnings.warn("Using less than 1000 runs. This may lead to unreliable results.")
 
     extract_mean = get_extract_mean(
-        scenario_config["manifold"], mtt_scenario=scenario_config["mtt"]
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded. The code change from this PR is already present on current main, and current main already has equivalent coverage in tests/test_evaluation_summarize_filter_results.py for accepting last_estimates without an explicit scenario_config['mtt'] flag. The only remaining branch delta is a duplicate test that now fails against the current summarize_filter_results API because it does not satisfy the required last_filter_states/last_estimates input shape.

PR #3629, opened 2026-07-01
Normalize PyTorch linspace dtype aliases

FlorianPfaff:
## Summary
- Normalize supported NumPy-style dtype aliases before calling PyTorch linspace.
- Preserve the existing integer dtype flooring behavior after dtype normalization.
- Add focused subprocess regression tests for public and raw PyTorch backend calls.

## Validation
- Checked that PyTorch rejects NumPy dtype aliases directly, while the PyRecEst dtype mapper resolves them to torch dtypes.
- Full suite not run in this connector-only environment.


FlorianPfaff:
Closing this older conflicted optional PyRecEst backend alias PR as low-value for now.

PR #3631, opened 2026-07-01
Fix summarize_filter_results last_estimates-only path

FlorianPfaff:
## Summary
- Accept `last_estimates` when `last_filter_states` is omitted in `summarize_filter_results`.
- Add a regression test covering estimate-only summaries.

## Bug fixed
`summarize_filter_results()` documents and errors as though either `last_filter_states` or `last_estimates` may be supplied. However, the control flow only accepted `last_estimates` when `last_filter_states` was also present, so valid estimate-only calls raised `ValueError`.

## Validation
- `python -m py_compile /tmp/summarize_filter_results.py /tmp/test_evaluation_summarize_filter_results.py`
- Verified branch diff against `main` (`ahead_by=2`, `behind_by=0`).
- Full test suite not run locally; the container cannot resolve `github.com` for cloning dependencies.

### DIFF
--- a/src/pyrecest/evaluation/summarize_filter_results.py
+++ b/src/pyrecest/evaluation/summarize_filter_results.py
@@ -27,6 +27,8 @@ def summarize_filter_results(
             "Provided both last_filter_states and last_estimates. Using last_estimates."
         )
         filter_results = last_estimates
+    elif last_estimates is not None:
+        filter_results = last_estimates
     elif last_filter_states is not None:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded. Current main already includes this behavior via PR #3660 (commit 94a4d559), including coverage for accepting last_estimates without last_filter_states and without an explicit mtt flag. Rebasing this branch would only duplicate that merged change.

PR #3633, opened 2026-07-01
Fix PyTorch comparison device contract

FlorianPfaff:
## Summary
- Extend the PyTorch backend support device-placement hook to cover `greater`, `less`, and `logical_or`.
- Prefer an existing non-CPU tensor device before delegating to the underlying helper, matching the existing logical/equality device contract behavior.
- Keep raw PyTorch backend and active public PyTorch backend helpers synchronized when the patch is applied.

## Bug fixed
The package initializer already wrapped `greater`, `less`, and `logical_or`, but that wrapper picked the first tensor operand's device. Calls with a CPU tensor on the left and a CUDA/meta tensor on the right could therefore try to move the right operand back to CPU or evaluate mixed-device operands, unlike the newer backend-support hooks that prefer existing non-CPU devices.

## Validation
- Syntax-checked the updated module content locally with `py_compile`.
- Smoke-checked the device-selection wrapper logic locally with PyTorch CPU-left/meta-right operands for `greater`, `less`, and `logical_or`.
- Full suite not run in this environment because direct repository cloning is unavailable.

### DIFF
--- a/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
+++ b/src/pyrecest/backend_support/_torch_dtype_promotion_contract/__init__.py
@@ -2,19 +2,19 @@
 
 from __future__ import annotations
 
-import importlib.util
+from importlib import import_module, util as importlib_util
 from pathlib import Path
 
 
 def _load_base_contract_module():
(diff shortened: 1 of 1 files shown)


FlorianPfaff:

PR #3636, opened 2026-07-01
Fix PyTorch matmul mixed-device operands

FlorianPfaff:
## Summary
- Add a PyTorch matmul compatibility hook that chooses an existing non-CPU tensor device before dtype promotion.
- Keep raw and public PyTorch backend matmul synchronized.
- Add focused subprocess regressions for CPU/array-like inputs combined with CUDA/meta tensors.

## Bug fixed
PyTorch matmul converted operands independently and only promoted dtype. Mixed CPU plus CUDA/meta operands could then reach torch.matmul on different devices and fail.

## Validation
- Syntax-checked the new hook and regression test with py_compile.
- Compared branch with main: ahead by 3 commits, behind by 0.
- Full suite not run in this connector session; CI should run the focused regressions.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_matmul_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_matmul_device_contract.py
@@ -1,0 +1,57 @@
+"""PyTorch ``matmul`` device compatibility hook."""
+
+from __future__ import annotations
+
+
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this conflicted PR as superseded by the main-branch PyTorch device compatibility hook repair in d4f7ba21.

PR #3637, opened 2026-07-01
Fix PyTorch maximum/minimum device placement

FlorianPfaff:
## Summary
- Add a PyTorch backend-support hook for `maximum` and `minimum` that prefers an existing non-CPU tensor device before evaluating the elementwise operation.
- Wire the hook into the normal package import path so both raw `pyrecest._backend.pytorch` helpers and the active public `pyrecest.backend` facade are patched.
- Add subprocess regression coverage for raw and public PyTorch backend calls using CUDA when available and `meta` otherwise.

## Bug fixed
`pyrecest._backend.pytorch.maximum` and `minimum` coerced their operands independently via `array(...)`. Calls combining array-like or CPU operands with an existing CUDA/meta tensor could therefore end up invoking `torch.maximum`/`torch.minimum` on mixed-device tensors. The new hook moves both operands to the preferred existing tensor device and promotes to a common dtype before calling PyTorch.

## Validation
- Inspected the branch diff against `main` (`ahead_by=3`, `behind_by=0`).
- Syntax-checked the new hook and regression test content locally with `ast.parse`.
- Smoke-checked PyTorch `meta` execution for `torch.maximum`/`torch.minimum` locally.
- Full repository test suite not run in this connector session; CI should run the focused regressions.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
@@ -1,0 +1,63 @@
+"""PyTorch ``maximum``/``minimum`` device compatibility hook."""
+
+from __future__ import annotations
+
+
+def _preferred_pytorch_device(torch_module, *values):
+    """Return a non-CPU tensor device when mixed-device operands are present."""
+    for value in values:
+        if torch_module.is_tensor(value) and value.device.type != "cpu":
+            return value.device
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this conflicted PR as superseded by the main-branch PyTorch device compatibility hook repair in d4f7ba21.

PR #3638, opened 2026-07-01
Fix PyTorch matmul device placement

FlorianPfaff:
## Summary
- Preserve an existing non-CPU PyTorch tensor device when `matmul` receives mixed CPU/accelerator operands.
- Keep raw `pyrecest._backend.pytorch.matmul` and the active public PyTorch backend facade synchronized.
- Preserve existing dtype promotion before delegating to `torch.matmul`.

## Bug fixed
`matmul` converted operands independently and then delegated to `torch.matmul`. A CPU tensor or array-like left operand paired with a CUDA/meta right operand could therefore run on or return a CPU tensor instead of preserving the existing accelerator operand.

## Validation
- Syntax-checked the updated backend-support module locally with `py_compile`.
- Smoke-checked the wrapper logic locally with CPU-left/meta-right tensors.
- Full repository test suite not run in this connector session.


FlorianPfaff:
Closing this as superseded. Current `main` already has the dedicated `backend_support/_pytorch_matmul_device_contract.py` hook plus real regression coverage in `tests/backend_support/test_pytorch_matmul_device_contract.py`. This older branch conflicts with that test file and carries a less complete inline wrapper/placeholder-test history.

PR #3639, opened 2026-07-01
Fix PyTorch binary scalar boxing device placement

FlorianPfaff:
## Summary
- Preserve an existing tensor placement when PyTorch binary scalar wrappers box Python/list operands.
- Move already boxed tensor operands to the selected placement before calling the underlying PyTorch binary operation.
- Add a backend regression covering the non-CPU tensor operand path.

## Bug fixed
The previous wrapper implementation materialized Python/list operands with plain `torch.tensor(...)`. That could put the newly boxed operand somewhere different from an existing tensor operand, causing binary operations such as `arctan2` to fail with a placement mismatch.

## Validation
- Syntax-checked the changed module and regression test locally with `py_compile`.
- Reproduced the old placement mismatch locally and verified that the patched boxing path returns a result with the expected shape and placement.
- Full repository suite not run locally because direct repository cloning is unavailable in this execution environment.

### DIFF
--- a/src/pyrecest/_backend/pytorch/_dtype.py
+++ b/src/pyrecest/_backend/pytorch/_dtype.py
@@ -107,6 +107,16 @@ def _dtype_as_str(dtype):
     return str(dtype).split(".")[-1]
 
 
+def _preferred_tensor_device(*values):
+    tensor_devices = [value.device for value in values if _torch.is_tensor(value)]
+    if not tensor_devices:
+        return None
+    return next(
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded. Current main now has the PyTorch binary device contract patch for arctan2/mod/power in backend_support/_torch_dtype_promotion_contract, including the same preferred existing non-CPU tensor device behavior. This branch now conflicts in the older _dtype.py helper path, and resolving it would duplicate the newer compatibility layer for little remaining value.

PR #3643, opened 2026-07-01
Fix PyTorch one_hot scalar labels

FlorianPfaff:
## Summary
- Add a PyTorch `one_hot` compatibility hook that treats scalar integer labels as label values instead of tensor sizes.
- Keep public and raw PyTorch backend `one_hot` synchronized after package import.
- Add PyTorch-specific regression coverage for scalar-label one-hot encoding.

## Bug fixed
The PyTorch backend converted non-tensor labels with `torch.LongTensor(labels)`. For scalar integer labels such as `1`, PyTorch interprets the integer as a tensor length and allocates an uninitialized label tensor, which can raise out-of-range errors or produce nondeterministic results. The new hook uses scalar-index normalization for scalar labels while preserving the existing sequence-label behavior.

## Testing
- Not run in this GitHub-only session; focused regression tests were added.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
@@ -1,0 +1,63 @@
+"""PyTorch ``one_hot`` scalar-label compatibility hook."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3644, opened 2026-07-01
Fix shared NumPy assignment_by_sum duplicate indices

FlorianPfaff:
## Summary
- Add a shared NumPy/autograd compatibility hook so `assignment_by_sum` accumulates repeated advanced indices instead of applying NumPy's buffered `+=` semantics.
- Load the hook during package initialization through the existing stability compatibility path.
- Add a focused backend-portable regression for duplicate vector and tuple indices.

## Bug fixed
For shared NumPy backends, `assignment_by_sum(x, [1, 2], [0, 0])` previously produced `[2, 0, 0]` because NumPy's advanced-indexed `+=` buffers repeated indices. PyTorch and JAX accumulation paths add both contributions and produce `[3, 0, 0]`. The hook uses `numpy.add.at` for the non-vectorized shared NumPy path to match true scatter-add semantics.

## Validation
- `python -m py_compile` on the new compatibility hook and regression test in a local scratch copy.
- Verified the GitHub branch diff against `main` after resetting an earlier too-broad file replacement.
- Full repository test suite not run in this connector-only environment; CI should run the focused backend-portable regression.

### DIFF
--- a/src/pyrecest/backend_support/_shared_numpy_assignment_by_sum_contract.py
+++ b/src/pyrecest/backend_support/_shared_numpy_assignment_by_sum_contract.py
@@ -1,0 +1,81 @@
+"""Shared NumPy ``assignment_by_sum`` duplicate-index compatibility hook."""
+
+from __future__ import annotations
+
+import sys
+
+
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3645, opened 2026-07-01
Fix PyTorch pad NumPy mode aliases

FlorianPfaff:
## Summary
- Add a PyTorch backend-support hook that maps NumPy pad mode names `edge` and `wrap` to PyTorch's `replicate` and `circular` modes.
- Load the hook through the existing stability/backend-support path so raw and active public PyTorch backends are patched.
- Add raw and public PyTorch regressions for `pad(..., mode="edge")` and `pad(..., mode="wrap")`.

## Bug fixed
The PyTorch backend exposes a NumPy-style `pad` API but forwarded non-constant `mode` values directly to `torch.nn.functional.pad`. Valid NumPy mode names like `edge` and `wrap` are rejected by PyTorch, which expects `replicate` and `circular`.

## Validation
- Syntax-compiled the new hook and regression test locally.
- Checked the underlying PyTorch `replicate`/`circular` outputs for the regression cases locally.
- Full suite not run in this connector/container environment because cloning the repository failed due unavailable github.com DNS resolution.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_pad_mode_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_pad_mode_contract.py
@@ -1,0 +1,47 @@
+"""PyTorch ``pad`` NumPy mode-name compatibility hook."""
+
+from __future__ import annotations
+
+_PAD_MODE_ALIASES = {
+    "edge": "replicate",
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3646, opened 2026-07-01
Support PyTorch triangular helpers for array-like inputs

FlorianPfaff:
## Summary
- Patch raw PyTorch `vec_to_diag`, `tril_to_vec`, and `triu_to_vec` so they coerce array-like inputs through the backend array constructor.
- Keep public PyTorch backend helpers synchronized when PyTorch is the active backend.
- Move generated triangular indices onto the input tensor device before indexing.

## Bug fixed
The raw PyTorch triangular-vector helpers assumed tensor inputs: `vec_to_diag([1.0, 2.0])` passed a Python list to `torch.diag_embed`, while `tril_to_vec([[...]])` and `triu_to_vec([[...]])` accessed `.shape` on a list. CUDA tensors could also be indexed with CPU-generated triangular indices. The wrapper normalizes inputs and index devices.

## Validation
- Inspected the branch diff against current `main`.
- Full test suite was not run because the repository could not be cloned in this environment.

### DIFF
--- a/src/pyrecest/stability.py
+++ b/src/pyrecest/stability.py
@@ -10,7 +10,68 @@ from pyrecest.backend_support._pytorch_allclose_device_contract import (
     patch_pytorch_allclose_device_contract as _patch_pytorch_allclose_device_contract,
 )
 
+
+def _patch_pytorch_triangular_vector_helpers() -> None:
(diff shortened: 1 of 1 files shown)


FlorianPfaff:

PR #3647, opened 2026-07-01
Guard stability status lookup against non-string names

FlorianPfaff:
## Summary
- make `get_public_api_status()` return `None` for non-string lookup names before dictionary access
- avoid `TypeError` for unhashable values such as lists
- add a regression test covering non-string lookup input

## Bug fixed
`get_public_api_status()` represented a public metadata lookup where unknown names return `None`, but passing an unhashable non-string value such as `["KalmanFilter"]` reached `_PUBLIC_API_STATUS.get(...)` and raised `TypeError`. The helper now treats non-string names as unknown lookup keys.

## Testing
- Not run locally: the execution environment could not resolve `github.com` for cloning/installing the repository.

### DIFF
--- a/src/pyrecest/stability.py
+++ b/src/pyrecest/stability.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Iterable
 from dataclasses import asdict, dataclass
-from typing import Final, Literal, ParamSpec, TypeVar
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this conflicted optional PyRecEst stability PR as low-value for now; current main already has direct stabilization fixes and this conflicts with the shared stability registry area.

PR #3657, opened 2026-07-02
Normalize manifold names for axis labels

FlorianPfaff:
## Summary
- Normalize `get_axis_label` inputs with `strip().lower()` before matching supported manifold names.
- Preserve the existing specific-before-generic matching order so bounded/symmetric manifolds are not shadowed.
- Add regression coverage for whitespace-padded and mixed-case supported names.

## Validation
- Ran a focused local syntax/behavior check for the patched `get_axis_label` cases.
- Inspected branch diff against `main`: two commits, two changed files, no unrelated changes.
- Full repository test suite not run locally because this sandbox cannot resolve `github.com` for cloning/installing the full repository.


FlorianPfaff:
Closing this conflicted optional PyRecEst axis-label PR as low-value for now; it conflicts after newer main stabilization and automerge activity.

PR #3680, opened 2026-07-02
Reject text hypertoroidal uniform sample counts

FlorianPfaff:
## Summary
- reject textual sample counts in `HypertoroidalUniformDistribution.sample`
- add a regression test that `sample("3")` raises instead of silently drawing three samples

## Rationale
`np.asarray("3").item()` produces a scalar string and the previous validation accepted it via `int()`/`float()` conversion. That is inconsistent with the stricter sample-count validation used elsewhere and lets an invalid API type pass silently.

## Testing
- Not run locally: the execution environment cannot clone GitHub directly, so I used the GitHub connector for the patch.


FlorianPfaff:
Closing this conflicted validation PR as low-value PyRecEst maintenance work for this sweep.

PR #3699, opened 2026-07-02
Correct backend dot behavior

FlorianPfaff:
Fixes a PyTorch backend dot edge case for matrix inputs and adds a regression test.


FlorianPfaff:
Closing this because it conflicts with the established PyRecEst backend dot contract. The branch changes 2-D PyTorch dot behavior to matrix multiplication, but current tests expect dot to use the project-wide last-axis inner-product semantics, e.g. matrix/matrix inputs produce one value per leading row. Making this PR green would require weakening that existing contract, so it is not worth carrying as-is.

PR #3713, opened 2026-07-02
Patch raw PyTorch round for array-like inputs

FlorianPfaff:
## Summary
- patch raw/public PyTorch `round` so it accepts NumPy-style array-like inputs even when the active public backend is NumPy
- preserve the `decimals` keyword and `out` behavior
- add a subprocess regression test that imports `pyrecest._backend.pytorch` while `PYRECEST_BACKEND=numpy`

## Testing
- Not run locally: the execution environment here cannot clone GitHub repositories or install the full project test environment.


FlorianPfaff:

PR #3721, opened 2026-07-03
Patch raw PyTorch reshape under NumPy public backend

FlorianPfaff:
## Summary
- add an import-time compatibility hook for raw `pyrecest._backend.pytorch.reshape`
- coerce array-like reshape inputs through the raw PyTorch backend rather than the active public backend
- cover the NumPy-public-backend case with a subprocess regression test

## Bug fixed
When `PYRECEST_BACKEND=numpy`, importing `pyrecest` left `pyrecest._backend.pytorch.reshape` as the raw `torch.reshape` function. Calls such as `raw_pytorch.reshape([1, 2, 3, 4], np.array([2, 2]))` therefore failed because `torch.reshape` does not accept Python lists or NumPy-style shape arrays directly. The new hook patches the raw PyTorch helper regardless of the selected public backend while only updating `pyrecest.backend.reshape` when the public backend is actually PyTorch.

## Testing
- Added `tests/backend_support/test_raw_pytorch_reshape_contract.py`
- Not run locally; this execution environment cannot clone the GitHub repository directly.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_raw_reshape_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_raw_reshape_contract.py
@@ -1,0 +1,69 @@
+"""PyTorch raw ``reshape`` compatibility hook."""
+
+from __future__ import annotations
+
+from operator import index as _operator_index
+
+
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3746, opened 2026-07-03
Fix shared vec_to_diag array-like inputs

FlorianPfaff:
## Summary
- patch the shared NumPy/Autograd `vec_to_diag` facade so Python lists and other array-like vectors are coerced before diagonalization
- update the public and raw active shared backend references consistently
- add regression coverage for vector and batched array-like inputs

## Tests
- `python -m py_compile /mnt/data/stability_new.py`

I could not run the full repository pytest suite locally because this environment could not clone GitHub over git; CI should exercise the branch.


FlorianPfaff:

PR #3781, opened 2026-07-04
Fix comparison backend contract

FlorianPfaff:


FlorianPfaff:
Closing this as superseded: its patch is already present upstream on `main` (verified with `git cherry` against `origin/main`), so keeping the PR open only leaves a stale branch and stale checks.

PR #3784, opened 2026-07-04
Fix JAX triangular vector helpers for array-like inputs

FlorianPfaff:
## Summary
- Patch raw/public JAX `tril_to_vec` and `triu_to_vec` during stability initialization so they coerce NumPy-style array-like matrices before reading shape metadata.
- Keep the active public JAX backend facade synchronized with the patched raw helpers.
- Add regression coverage for raw JAX under the default public backend and the selected public JAX backend.

## Bug fixed
`tril_to_vec([[...]])` and `triu_to_vec([[...]])` in the JAX backend could access `.shape` before backend array coercion, so Python lists and other array-like matrix inputs failed instead of following the backend's NumPy-style array-like contract.

## Validation
- Confirmed the branch is based on current `main` with `compare_commits` (`behind_by=0`).
- Syntax-checked the modified `stability.py` and new test content locally via Python `compile()`.
- Reproduced the underlying JAX failure mode in the sandbox: the pre-fix logic raises `AttributeError` on a Python list, while coercing with `jnp.asarray` returns the expected triangular vector.
- Full repository pytest was not run locally because this environment could not clone the repository directly; the PR adds targeted tests for CI.

### DIFF
--- a/src/pyrecest/stability.py
+++ b/src/pyrecest/stability.py
@@ -246,6 +246,66 @@ def _patch_jax_squeeze_numpy_contract() -> None:
         backend.squeeze = squeeze
 
 
+def _patch_jax_triangular_vec_numpy_contract() -> None:
+    """Patch raw/public JAX triangular vector helpers for array-like inputs."""
+    try:
+        import jax.numpy as jnp  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3785. That PR carries the broader matrix-helper array-like backend coverage, including the JAX triangular-vector helper behavior this PR targets.

PR #3788, opened 2026-07-04
Fix PyTorch linspace device contract

FlorianPfaff:
## Summary
- Extend the existing PyTorch device compatibility hook to patch `linspace`.
- Prefer an existing non-CPU endpoint device before calling the raw PyTorch helper.
- Add regression coverage for raw PyTorch after importing `pyrecest` and for the public PyTorch backend.

## Validation
- Syntax-checked the patched module and test file locally.
- Ran targeted raw/public PyTorch smoke checks with a meta endpoint in a temporary source tree.
- Confirmed the branch is current with `main` (`behind_by=0`).

Full repository pytest was not run locally; CI should run the new backend-portable tests.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
@@ -1,4 +1,4 @@
-"""PyTorch ``maximum``/``minimum`` device compatibility hook."""
+"""PyTorch device compatibility hooks."""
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3810, which keeps the same PyTorch linspace device contract fix and adds the newer integer-num behavior preservation. Keeping both open would duplicate the same source/test changes and CI.

PR #3792, opened 2026-07-04
Activate PyTorch minmax device patch on import

FlorianPfaff:
## Summary
- activate the existing PyTorch `maximum`/`minimum` device compatibility patch during the normal `import pyrecest` path
- keeps array-like operands on an existing non-CPU tensor device, matching the existing regression coverage for raw and public PyTorch backends

## Testing
- Not run locally; change is covered by the existing `tests/backend_support/test_pytorch_minmax_device_contract.py` regression test.

### DIFF
--- a/src/pyrecest/evidence.py
+++ b/src/pyrecest/evidence.py
@@ -14,6 +14,12 @@ from typing import Any, Literal
 
 import numpy as np
 
+from pyrecest.backend_support._pytorch_minmax_device_contract import (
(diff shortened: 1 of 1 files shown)


FlorianPfaff:
Closing this as superseded and misdirected: the PR title is about activating the PyTorch minmax device patch, but the diff only imports that patch from `evidence.py`. The minmax device/import support is already handled on `main` by #3759, and adding this import side effect from `evidence.py` is not worth carrying.

PR #3802, opened 2026-07-04
Fix PyTorch comparison equality contract

FlorianPfaff:
## Summary
- Patch PyTorch comparison helpers for `equal` and `less_equal` in the backend compatibility layer.
- Keep raw PyTorch and the active public PyTorch backend facade synchronized.
- Preserve mixed-device handling for comparison operands.

## Validation
- Inspected the branch diff against `main`: one commit, one modified file, +11/-1.
- Full local pytest was not run in this connector-only environment.


FlorianPfaff:
Closing this as already applied upstream. Rebasing the branch onto current main skipped its commit as already present, leaving no remaining diff for this PR.

PR #3804, opened 2026-07-04
Fix PyTorch clamp device handling

FlorianPfaff:
Fixes a PyTorch backend bounded-value helper edge case and adds focused regression tests.

Validation: branch is current with main. An isolated PyTorch smoke check was run. Full repository tests were not run here.


FlorianPfaff:

PR #3810, opened 2026-07-04
Fix PyTorch linspace device selection

FlorianPfaff:
## Summary
- patch PyTorch `linspace` to prefer an existing non-CPU endpoint device, including `meta`, instead of always using the first tensor endpoint's device
- preserve the previous dtype promotion, broadcasting, `endpoint`, and integer-`num` behavior
- add raw/public PyTorch regression tests that use CUDA when available and `meta` as a deterministic fallback

## Tests
- Verified the patched `linspace` logic locally with PyTorch meta tensors
- Added `tests/backend_support/test_pytorch_linspace_device_contract.py`


FlorianPfaff:

PR #3811, opened 2026-07-04
Fix PyTorch array sequence device handling

FlorianPfaff:
## Summary
- Teach the PyTorch `_common.array()` helper to scan nested sequence inputs for an existing non-CPU tensor device before recursively stacking sequence entries.
- Move array-like and CPU sequence elements to that preferred existing device so `torch.stack` receives same-device tensors.
- Add regression coverage for public and raw PyTorch backends using a `meta` tensor.

## Bug fixed
`pyrecest._backend.pytorch.array(([1.0, 2.0], torch.empty((2,), device="meta")))` materialized the list element on CPU while leaving the tensor element on `meta`, then failed in `torch.stack` with a mixed-device runtime error. The patch preserves the existing non-CPU tensor device for flat and nested sequence construction.

## Validation
- Reproduced the old mixed CPU/meta `torch.stack` failure in an isolated torch-only check.
- Syntax-checked the updated backend helper and regression test with `python -m py_compile`.
- Ran an isolated torch check of the patched helper for flat and nested sequence inputs, verifying `meta`-device outputs.
- Compared branch against `main`: 2 commits ahead, 0 behind; 2 files changed, +62/-1.

I did not run the full PyRecEst pytest suite in this connector-only environment.

### DIFF
--- a/src/pyrecest/_backend/pytorch/_common.py
+++ b/src/pyrecest/_backend/pytorch/_common.py
@@ -52,6 +52,31 @@ def _as_torch_compatible_numpy_array(x):
     return x
 
 
+def _preferred_non_cpu_tensor_device(value):
+    if _torch_is_tensor(value):
+        return value.device if value.device.type != "cpu" else None
+    if isinstance(value, (list, tuple)):
+        for item in value:
+            device = _preferred_non_cpu_tensor_device(item)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3812, opened 2026-07-04
Fix JAX one_hot bounds handling

FlorianPfaff:
## Summary
- Patch the runtime one_hot compatibility hook so raw and public JAX one_hot return zero rows for labels outside `[0, num_classes)` instead of indexing the final class.
- Preserve the existing uint8 output and labels/num_classes keyword contract.
- Add backend-portable regressions for public JAX and raw JAX after package import.

## Bug fixed
The JAX backend one_hot compatibility path used identity-matrix indexing. With JAX indexing, labels such as `-1` or `num_classes` can select the final row, producing a false final-class activation. The patch computes a validity mask, indexes only safe labels, and zeros rows whose labels are outside the class range.

## Validation
- `python -m py_compile /mnt/data/_pytorch_one_hot_scalar_contract.py /mnt/data/test_jax_one_hot_contract.py`
- Local JAX smoke check for `[-1, 0, 2, 3]` with `num_classes=3` returning zero rows for out-of-range labels and preserving `uint8` dtype.
- Full pytest was not run locally because this environment cannot clone GitHub repositories via DNS; CI should run the focused backend-portable regressions.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from operator import index as _operator_index
 
 
-def patch_pytorch_one_hot_scalar_contract() -> None:
+def _patch_pytorch_one_hot_scalar_contract() -> None:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3828, opened 2026-07-05

FlorianPfaff:
## Summary
- normalize PyTorch `linalg.logm` inputs through the existing linalg tensor coercion helper before invoking the autograd-backed implementation
- preserve the existing `_Logm` autograd path for tensor inputs
- add focused raw/public PyTorch regression coverage for integer Python-list matrix inputs

## Bug fixed
`pyrecest._backend.pytorch.linalg.logm` exposed `_Logm.apply` directly, unlike neighboring linalg helpers that first call `_as_linalg_tensor`. As a result, calls such as `linalg.logm([[1, 0], [0, 1]])` could pass a Python list into `ctx.save_for_backward`, failing before the SciPy bridge and contradicting PyRecEst's NumPy-style array-like backend contract. Integer array-like inputs also skipped the standard float/complex linalg promotion.

## Validation
- `python -m py_compile /tmp/linalg_new.py /tmp/test_pytorch_linalg_logm_arraylike_contract.py`
- local minimal PyTorch/SciPy stub smoke test: `logm([[1, 0], [0, 1]])` returns a floating zero matrix and tensor-input gradients still backpropagate
- `compare_commits`: branch is 2 commits ahead of `main`, 0 behind; changed files are limited to `src/pyrecest/_backend/pytorch/linalg.py` and the new regression test

Full repository pytest was not run in this connector environment; CI should run the added focused regression.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -236,7 +236,9 @@ class _Logm(_torch.autograd.Function):
         return _Logm._logm(backward_tensor).to(tensor.dtype)[..., :n, n:]
 
 
-logm = _Logm.apply
+def logm(x):
+    """Compute matrix logarithm after array-like input promotion."""
+    return _Logm.apply(_as_linalg_tensor(x))
 
 
 def sqrtm(x):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3847, which carries the same PyTorch logm array-like fix with newer focused regression coverage. Keeping one active PR avoids duplicate CI and review.

PR #3829, opened 2026-07-05
Reject None in time offset numeric arrays

FlorianPfaff:
## Summary
- Reject `None` entries in object-dtype time/value arrays before NumPy can coerce them to `NaN`.
- Keep the existing validation behavior for bool/text object payloads and numeric object arrays.
- Add focused regression tests for `apply_time_offset`, `nearest_time_indices`, `interpolate_reference_values`, and `time_offset_error_summary`.

## Validation
- Added focused regression tests.
- Full suite not run in this connector-only environment; CI should run on the PR.

## Notes
- I also found a separate PyTorch top-level `pyrecest.copy` export synchronization issue, but the connector blocked the larger backend-hook update, so this PR intentionally fixes the smaller calibration bug without mixing concerns.

### DIFF
--- a/src/pyrecest/calibration/time_offset.py
+++ b/src/pyrecest/calibration/time_offset.py
@@ -25,8 +25,17 @@ class TimeOffsetFitResult:
     metadata: Mapping[str, Any] = field(default_factory=dict)
 
     def summary(self) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded. Current main already includes the None-value validation for time-offset numeric inputs via PR #3767 (commit 911b4737), so this conflicting duplicate branch is no longer needed.

PR #3830, opened 2026-07-05
Reject bool inputs in PyTorch one_hot

FlorianPfaff:
## Summary
- add a small guard before the PyTorch one_hot scalar shim coerces inputs to torch.long
- keep integer scalar, sequence, and tensor inputs on the existing uint8 one_hot path
- keep raw PyTorch and active public PyTorch facades synchronized

## Bug fixed
The scalar shim already avoided converting bool tensors, but Python and NumPy bool inputs still reached torch.as_tensor(..., dtype=torch.long), so True/False could be treated as class indices. This change rejects those inputs consistently before conversion.

## Validation
- python -m py_compile /tmp/_pytorch_one_hot_scalar_contract.py /tmp/test_pytorch_one_hot_scalar_contract.py
- local smoke check of the new bool-input detector with Python, NumPy, and torch values

Note: I attempted to add focused regression coverage under tests/backend_support, but the connector safety layer rejected writes to that test path in this session. The implementation change is intentionally limited to the compatibility shim.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_one_hot_scalar_contract.py
@@ -5,6 +5,24 @@ from __future__ import annotations
 from operator import index as _operator_index
 
 
+def _is_numpy_bool_scalar(value) -> bool:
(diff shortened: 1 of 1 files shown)


FlorianPfaff:

PR #3831, opened 2026-07-05

FlorianPfaff:
## Summary
- Normalize PyTorch `linalg.logm` inputs through the existing linalg tensor coercion helper.
- Keep the differentiable tensor path intact while accepting Python-list and integer array-like matrices.
- Add regression coverage for the raw PyTorch linalg module and the public PyTorch backend path.

## Validation
- Syntax-checked the modified Python content locally with `compile()`.
- Full repository tests were not run locally; CI should run the new focused regression test.


FlorianPfaff:
Closing as superseded. Current main already includes the PyTorch linalg.logm array-like fix via PR #3833 (commit f591a9d2), so rebasing this conflicting duplicate branch would not add useful new behavior.

PR #3832, opened 2026-07-05
Sync top-level copy export for PyTorch backend

FlorianPfaff:
## Summary
- refresh the package-level `pyrecest.copy` export after import-time backend facade patches run
- add a PyTorch backend-portable regression asserting `pyrecest.copy`, `pyrecest.backend.copy`, and raw PyTorch `copy` stay synchronized
- preserve existing raw PyTorch array-like copy coverage

## Bug fixed
With `PYRECEST_BACKEND=pytorch`, `pyrecest.__init__` imports `copy` before backend-support hooks patch the active backend facade. The raw and active backend `copy` helpers can therefore be patched while the package-level `pyrecest.copy` name remains bound to the pre-patch function.

## Validation
- Syntax-checked the modified module and regression test locally with `python -m py_compile`.
- Compared `fix-pytorch-copy-export` against `main`: `ahead_by=2`, `behind_by=0`, changed only `src/pyrecest/backend_tools.py` and `tests/backend_support/test_pytorch_copy_contract.py`.
- Full repository tests were not run locally because this sandbox cannot clone/install the repository from GitHub.

### DIFF
--- a/src/pyrecest/backend_tools.py
+++ b/src/pyrecest/backend_tools.py
@@ -3,8 +3,22 @@
 from __future__ import annotations
 
 import os
+import sys
 import warnings
 from operator import index as _operator_index
+
+
+def _sync_top_level_backend_copy_export() -> None:
+    """Refresh ``pyrecest.copy`` after import-time backend facade patches."""
+
+    pyrecest_module = sys.modules.get("pyrecest")
+    if pyrecest_module is None:
+        return
+    try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3837. This branch now conflicts with current main, and #3837 carries the same top-level PyTorch copy export synchronization in a cleaner backend-support hook with current-base tests. Keeping both open would duplicate review and CI.

PR #3838, opened 2026-07-05

FlorianPfaff:
## Summary
- wrap the PyTorch `linalg.logm` entry point so inputs pass through `_as_linalg_tensor` before reaching the custom autograd function
- add raw/public PyTorch regression coverage for Python-list integer matrix inputs

## Bug fixed
`logm = _Logm.apply` bypassed the linalg backend's normal array-like promotion path. That meant calls such as `pyrecest._backend.pytorch.linalg.logm([[1, 0], [0, 1]])` passed a Python list directly into `torch.autograd.Function.apply`, rather than accepting NumPy-style array-like inputs like the neighboring PyTorch linalg helpers.

## Validation
- Compared branch against `main`: 2 commits ahead, 0 behind; changes limited to `src/pyrecest/_backend/pytorch/linalg.py` and the new focused regression test.
- Syntax-checked the modified linalg module content locally with Python AST parsing.
- Full pytest suite not run in this connector-only environment; CI should run the new focused backend test.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -236,7 +236,9 @@ class _Logm(_torch.autograd.Function):
         return _Logm._logm(backward_tensor).to(tensor.dtype)[..., :n, n:]
 
 
-logm = _Logm.apply
+def logm(x):
+    """Compute a matrix logarithm after array-like input promotion."""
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3847, which carries the same PyTorch logm array-like fix with newer focused regression coverage. Keeping one active PR avoids duplicate CI and review.

PR #3842, opened 2026-07-05
Fix PyTorch unique return flags

FlorianPfaff:
## Summary
- patch raw/public PyTorch `unique` to accept NumPy-style `return_inverse` and `return_counts` flags
- keep array-like input coercion through the PyTorch backend helper
- add subprocess regression coverage for raw PyTorch under the NumPy public backend and for the selected public PyTorch backend

## Bug fixed
The PyTorch backend exposed `unique(ar, axis=None)` only, even though the underlying Torch helper supports inverse/count return values and the PyRecEst backend facade follows NumPy-style array APIs. Calls such as `backend.unique([2, 1, 2], return_counts=True)` therefore failed under `PYRECEST_BACKEND=pytorch` with an unexpected keyword argument instead of returning the requested counts.

## Validation
- Syntax-checked the modified hook and new regression test locally with Python `compile()`.
- Compared branch against `main`: 2 commits ahead, 0 behind; changes are limited to the PyTorch compatibility hook and the focused regression test.

Full pytest was not run in this connector-only environment; CI should run the added backend-support regression.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_allclose_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_allclose_device_contract.py
@@ -66,6 +66,7 @@ def _patch_pytorch_linalg_logm_arraylike_contract() -> None:
 
 def _patch_pytorch_flip_numpy_axis_contract() -> None:
     """Patch raw/public PyTorch ``flip`` to accept NumPy integer axes."""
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3847, opened 2026-07-05

FlorianPfaff:
## Summary
- wrap the PyTorch `linalg.logm` autograd function in the existing `_as_linalg_tensor` normalizer
- preserve the existing differentiable `_Logm` implementation for tensor inputs
- add a backend-portable PyTorch regression test for raw and public `logm` calls with Python-list matrix inputs

## Bug fixed
`pyrecest._backend.pytorch.linalg.logm` was exposed directly as `_Logm.apply`, so array-like or integer matrix inputs bypassed the PyTorch linalg input normalizer. Calls such as `logm([[1, 0], [0, 1]])` could fail before reaching the SciPy bridge even though neighboring PyTorch linalg helpers accept NumPy-style array-like inputs.

## Validation
- `python -m py_compile /tmp/linalg_new.py /tmp/test_pytorch_logm_arraylike_contract.py`
- Verified branch comparison against `main`: 2 commits ahead, 0 behind; changes limited to `src/pyrecest/_backend/pytorch/linalg.py` and `tests/backend_support/test_pytorch_logm_arraylike_contract.py`.

Full repository tests were not run locally because direct GitHub cloning is unavailable in this environment; CI should run the focused regression.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -236,7 +236,9 @@ class _Logm(_torch.autograd.Function):
         return _Logm._logm(backward_tensor).to(tensor.dtype)[..., :n, n:]
 
 
-logm = _Logm.apply
+def logm(x):
+    """Compute a matrix logarithm after PyRecEst-style array-like promotion."""
+    return _Logm.apply(_as_linalg_tensor(x))
 
 
 def sqrtm(x):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #3852 plus the regression coverage already merged in #3833. #3852 keeps the direct PyTorch linalg.logm implementation change without adding duplicate same-purpose tests.

PR #3853, opened 2026-07-05
Fix hypersphere subset Dirac marginalization

FlorianPfaff:
## Summary
- implement `marginalize_linear` for `LinHypersphereSubsetCartProdDiracDistribution` so the exported Cartesian-product Dirac class is instantiable and splits hypersphere/linear coordinates correctly
- coerce constructor samples with backend `asarray` before shape-based dimension bookkeeping, matching sibling Cartesian-product Dirac constructors
- add a focused regression covering array-like constructor inputs and both marginalizations

## Bug fixed
`LinHypersphereSubsetCartProdDiracDistribution` inherits the abstract `marginalize_linear` contract from `LinBoundedCartProdDiracDistribution`, but did not implement it. Instantiating the public class therefore failed as an abstract class instead of producing a hypersphere/linear product Dirac. The constructor also read `d.shape` before backend coercion, so list inputs could not satisfy the same array-like constructor contract as `HypercylindricalDiracDistribution`.

## Validation
- Syntax-checked the patched source and regression test text with Python `compile()`.
- Compared the branch against `main`: 2 commits ahead, 0 behind; changes are limited to the class implementation and focused regression test.
- Full pytest was not run in this sandbox/connector environment.

### DIFF
--- a/src/pyrecest/distributions/cart_prod/lin_hypersphere_subset_dirac_distribution.py
+++ b/src/pyrecest/distributions/cart_prod/lin_hypersphere_subset_dirac_distribution.py
@@ -1,8 +1,11 @@
 from typing import Union
 
 # pylint: disable=no-name-in-module,no-member
-from pyrecest.backend import int32, int64
+from pyrecest.backend import asarray, int32, int64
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3862, opened 2026-07-05
Fix PyTorch stack and concatenate device contract

FlorianPfaff:
## Summary
- Extend the existing PyTorch stack-helper device compatibility hook to cover `stack` and `concatenate`, not only hstack/vstack/column_stack/dstack.
- Normalize all sequence operands onto a preferred existing tensor device before dispatching to `torch.stack`/`torch.cat`.
- Preserve the previous dtype-widening behavior for normalized tensor sequences.
- Add subprocess regressions for both raw PyTorch helpers under the NumPy public backend and the active public PyTorch backend.

## Bug fixed
`pyrecest._backend.pytorch.stack` / `concatenate` and the public PyTorch backend facade still used the raw tensor sequence conversion path. Calls mixing NumPy-style array-like operands with an existing non-CPU or symbolic tensor, e.g. a `meta` tensor, could therefore call Torch with tensors on different devices and fail. The neighboring stack helpers already had device normalization; this applies the same contract to the generic sequence helpers.

## Validation
- Syntax-checked the modified hook and new regression test locally with `python -m py_compile`.
- Full pytest was not run locally because the execution environment cannot clone GitHub; CI should run the added backend-support subprocess tests.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_stack_helpers_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_stack_helpers_device_contract.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import importlib
 
 
-_STACK_HELPER_NAMES = ("hstack", "vstack", "column_stack", "dstack")
+_STACK_HELPER_NAMES = ("concatenate", "stack", "hstack", "vstack", "column_stack", "dstack")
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3864, opened 2026-07-05
Fix JAX one_hot out-of-range labels

FlorianPfaff:
Summary:
- add a dedicated JAX one_hot bounds compatibility hook
- wire it into the existing runtime backend patches
- add public and raw JAX regressions for labels outside [0, num_classes)

Bug fixed:
The JAX one_hot compatibility path used identity-matrix indexing directly. With JAX indexing, labels such as -1 or num_classes can select the final class instead of producing a zero row.

Validation:
- branch compare: 3 commits ahead, 0 behind main
- local syntax compilation of the edited Python content before committing

This supersedes the older non-mergeable PR #3812.


FlorianPfaff:

PR #3874, opened 2026-07-06
Fix PyTorch linalg norm boolean axis validation

FlorianPfaff:
Summary:
- reject boolean scalar axes in the PyTorch linalg.norm axis normalizer
- preserve NumPy integer-scalar axis support
- add PyTorch backend regression coverage for boolean and NumPy scalar axes

Bug fixed:
The PyTorch linalg.norm backend accepted boolean axes because bool is an int subclass in Python. That made axis=True silently select axis 1 instead of rejecting the non-integer axis contract.

Validation:
- Added tests in tests/backend_support/test_pytorch_linalg_norm_axis_contract.py
- Not run locally; changes were applied through the GitHub connector.


FlorianPfaff:
Closing this conflicted PyTorch backend cleanup as low-value for now; it still has failing PyTorch checks and cannot be updated cleanly from main.

PR #3877, opened 2026-07-06
Fix PyTorch random helper device selection

FlorianPfaff:
## Summary
- add a PyTorch random-device compatibility hook that prefers an existing non-CPU tensor device before falling back to CPU
- wire the hook into the existing runtime backend compatibility path
- add focused regression coverage for random helper device selection and a CUDA smoke path for `uniform` when available

## Bug fixed
The PyTorch random backend had several helper paths that selected the first tensor operand's device. With a CPU tensor first and a CUDA/MPS/meta tensor later, helpers such as `_normal_device`, `_tensor_device`, and `_randint_device` could choose CPU and either move data off the accelerator or fail when the later tensor cannot be materialized on CPU. The new hook mirrors the device-selection contract already used by PyTorch binary backend helpers.

## Validation
- Inspected the branch diff with the GitHub connector: 3 commits ahead of `main`, 0 behind.
- Added focused tests in `tests/backend_support/test_pytorch_random_device_contract.py`.
- Full pytest was not run in this connector-only environment; CI should run the backend matrix.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
@@ -4,6 +4,10 @@ from __future__ import annotations
 
 import importlib
 
+from pyrecest.backend_support._pytorch_random_device_contract import (
+    patch_pytorch_random_device_contract as _patch_pytorch_random_device_contract,
+)
+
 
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this older conflicted branch because the same PyTorch random-device fix is represented by newer PR #3887 with automerge enabled.

PR #3879, opened 2026-07-06
Reject boolean PyTorch matrix_power exponents

FlorianPfaff:
## Summary
- reject Python, NumPy, and Torch boolean scalar exponents in the PyTorch `linalg.matrix_power` wrapper
- preserve integer scalar exponents including NumPy integer scalars and zero-dimensional integer arrays
- add subprocess regression coverage for the public PyTorch backend

## Bug fixed
`bool` values implement Python integer-index semantics, so the previous PyTorch wrapper accepted `n=True` as exponent `1` and `n=False` as exponent `0`. Boolean scalars are not valid matrix-power exponents under the backend contract and should be rejected instead of silently changing the requested power.

## Validation
- Compared branch against `main`: 2 commits ahead, 0 behind.
- Local tests were not run in this environment; CI should exercise the added PyTorch-backend regressions.

### DIFF
--- a/src/pyrecest/_backend/pytorch/linalg.py
+++ b/src/pyrecest/_backend/pytorch/linalg.py
@@ -66,8 +66,21 @@ def _as_linalg_tensor(value):
     return tensor
 
 
+def _is_boolean_scalar(value):
+    """Return whether value is a scalar boolean accepted by Python indexing."""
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3882, opened 2026-07-06
Fix cartesian hypertoroidal grid regeneration shape

FlorianPfaff:
## Summary
- preserve full per-axis cartesian-product resolution in `HypertoroidalGridDistribution.n_grid_points`
- add a regression for cartesian-product hypertoroidal grids constructed without stored coordinates

## Bug fixed
For `grid_type="cartesian_prod"`, `get_grid()` regenerates coordinates from `self.n_grid_points` when `self.grid` is not stored. The inherited `AbstractGridDistribution.n_grid_points` returned only `grid_values.shape[0]`, so a 2-D grid with values shaped `(2, 3)` regenerated as a one-dimensional 2-point grid instead of a six-point 2-D cartesian product.

## Validation
- Compared the branch against `main`: 2 commits ahead, 0 behind; changes are limited to the hypertoroidal grid distribution and its focused regression test.
- Local pytest was not run because this environment cannot clone/install the repository; CI should run `tests/distributions/test_hypertoroidal_grid_distribution.py`.

### DIFF
--- a/src/pyrecest/distributions/hypertorus/hypertoroidal_grid_distribution.py
+++ b/src/pyrecest/distributions/hypertorus/hypertoroidal_grid_distribution.py
@@ -181,6 +181,13 @@ class HypertoroidalGridDistribution(
     def get_manifold_size(self):
         return AbstractHypertoroidalDistribution.get_manifold_size(self)
 
+    @property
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3884, opened 2026-07-06
Reject boolean PyTorch reduction axes

FlorianPfaff:
## Summary
- patch the PyTorch backend reduction-axis normalization path to reject scalar boolean axis/dim values before PyTorch can reinterpret them as integers
- cover `sum`, `mean`, `std`, arg reductions, boolean reductions, `count_nonzero`, and `max`
- keep valid integer-axis reductions unchanged

## Rationale
NumPy-style reductions generally reject boolean axis specifications. The PyTorch backend previously accepted values such as `axis=True` and silently treated them as dimension `1`, producing plausible but incorrect results.

## Follow-up note
Tuple/list axes for helpers that bypass `_normalize_reduction_axes` should be handled in a separate direct backend change; this PR keeps the regression focused on the scalar `axis=True`/`dim=True` behavior that caused silent mis-reductions across core reductions.

## Testing
- Added `tests/backend_support/test_pytorch_reduction_bool_axis_contract.py`
- CI requested on the PR branch; local repository tests were not run because this environment cannot clone/install the repository from GitHub.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_allclose_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_allclose_device_contract.py
@@ -229,6 +229,70 @@ def _patch_broadcast_arrays(pytorch_backend, backend, torch_module) -> None:
         backend.broadcast_arrays = broadcast_arrays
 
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3885, opened 2026-07-06
Normalize PyTorch diagonal NumPy scalar axes

FlorianPfaff:
## Summary
- normalize zero-dimensional NumPy integer arrays before forwarding PyTorch diagonal offset/axis arguments to `torch.diagonal`
- add a PyTorch backend regression for `np.array` scalar offset/axis arguments

## Bug fixed
The PyTorch backend uses the common `diagonal` helper. Its integer normalizer only tried `operator.index`, so NumPy zero-dimensional integer arrays such as `np.array(0)` were passed through to `torch.diagonal`, which rejects them. NumPy accepts these scalar integer arrays for `offset`, `axis1`, and `axis2`, so the backend failed a NumPy-style argument contract.

## Testing
- Added `tests/backend_support/test_pytorch_diagonal_numpy_scalar_axes.py`
- Not run locally; this environment cannot clone/install the repository from GitHub. CI should exercise the focused regression.

### DIFF
--- a/src/pyrecest/_backend/_common.py
+++ b/src/pyrecest/_backend/_common.py
@@ -45,6 +45,16 @@ def _normalize_size_axis(axis):
 
 def _normalize_diagonal_integer(value):
     """Return NumPy-style integer scalar arguments for stricter tensor APIs."""
+    if isinstance(value, _AXIS_FLAG_TYPES):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #3887, opened 2026-07-06
Fix PyTorch random helper device selection

FlorianPfaff:
## Summary
- add a PyTorch random-device compatibility hook that prefers existing non-CPU/meta tensor devices before falling back to CPU
- activate the hook through the existing PyTorch backend-support path
- keep the active public PyTorch `backend.random.uniform` facade synchronized with the patched raw helper
- add focused regression coverage for raw helper device selection and public-facade synchronization

## Bug fixed
The PyTorch random backend selected the first tensor operand's device in helper paths such as `_normal_device`, `_tensor_device`, and `_randint_device`. With a CPU tensor first and a CUDA/MPS/meta tensor later, those helpers could choose CPU and either move data off the accelerator or fail for tensors that cannot be materialized on CPU. The new hook mirrors the non-CPU preference used by other PyTorch backend compatibility helpers.

## Validation
- Compared branch against `main`: 3 commits ahead, 0 behind; changes limited to the PyTorch backend-support hook and focused tests.
- Did not run the full pytest suite locally because this connector session cannot clone/install the repository; CI should run the added regression test.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_minmax_device_contract.py
@@ -7,6 +7,10 @@ from operator import index as _operator_index
 
 import numpy as _np
 
+from pyrecest.backend_support._pytorch_random_device_contract import (
+    patch_pytorch_random_device_contract as _patch_pytorch_random_device_contract,
+)
+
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3888, opened 2026-07-06
Argsort validation update

FlorianPfaff:


FlorianPfaff:
Summary: rejects boolean axis/dim values in the PyTorch argsort compatibility wrapper, preserves valid scalar integer axes, and adds focused public/raw backend regression tests.

Validation: syntax-checked the changed files locally. Full test suite was not run in this connector environment.

PR #3890, opened 2026-07-06
Fix shared NumPy triangular vector helpers for rectangular inputs

FlorianPfaff:
## Summary

- fix NumPy and Autograd `tril_to_vec` / `triu_to_vec` to use both trailing matrix dimensions for rectangular inputs
- coerce array-like inputs before triangular indexing
- add subprocess regression coverage for NumPy and Autograd backends

## Bug fixed

The shared NumPy triangular vector helpers derived triangular indices from `x.shape[-1]` only. For rectangular matrices this either missed valid entries or produced row indices outside the number of rows, for example `tril_to_vec([[1, 2, 3], [4, 5, 6]])` attempted to index row 2 of a 2-row matrix. The backend-specific overrides now call `tril_indices` / `triu_indices` with explicit row and column dimensions.

## Validation

- Added `tests/backend_support/test_shared_numpy_triangular_helpers_contract.py`
- Verified the branch diff contains only the NumPy/Autograd helper overrides and the focused regression test via the GitHub connector
- Full test suite was not run locally because this execution environment cannot clone/install the repository from GitHub

### DIFF
--- a/src/pyrecest/_backend/autograd/__init__.py
+++ b/src/pyrecest/_backend/autograd/__init__.py
@@ -1,5 +1,7 @@
 """Autograd based computation backend."""
 
+from operator import index as _operator_index
+
 import autograd.numpy as _np
 from autograd.numpy import (
     all,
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #3910, opened 2026-07-06
Validate NumPy choice population size

FlorianPfaff:
## Summary
- add an explicit NumPy backend `random.choice` population-size validation wrapper
- reject non-positive scalar populations and empty array populations even when `size=0`
- cover scalar, array-axis, and non-population empty-dimension cases with regression tests

## Rationale
`random.choice` already reports that `a` must be a positive integer or a non-empty array for boolean scalar populations. However, NumPy permits empty draws from empty or non-positive populations when `size=0`, so PyRecEst accepted invalid populations in that edge case. This patch makes validation independent of the requested sample shape and keeps empty dimensions that are not the sampled population axis valid.

## Testing
- Not run locally; patch was prepared through the GitHub connector.

### DIFF
--- a/src/pyrecest/_backend/numpy/random.py
+++ b/src/pyrecest/_backend/numpy/random.py
@@ -1,5 +1,7 @@
 """Numpy based random backend."""
 
+from operator import index as _operator_index
+
 import numpy as _np
 from numpy.random import default_rng as _default_rng
 from numpy.random import (  # For PyRecEst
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this for now. The PR changes the shared random.choice behavior to reject zero-sized draws from nonpositive populations, but it only implements the NumPy side; JAX/PyTorch still fail the updated shared backend test. This also overlaps with #3914, where the backend-wide implementation shows the same behavior-policy conflict with existing tests, so this is not an easy cleanup merge.

PR #3914, opened 2026-07-06
Validate backend random.choice populations

FlorianPfaff:
## Summary
- add a backend-support patch that validates PyTorch and JAX `random.choice` populations before sampling
- wire the patch through the existing random backend compatibility hook
- add focused regression tests for non-positive scalar populations and empty array populations with zero requested samples

## Bug fixed
The PyTorch and JAX `random.choice` backends could return an empty result for invalid populations when `size` contained zero samples, e.g. `choice(0, size=0)` or choosing from an empty population axis with `size=0`. Population validity should not depend on the requested output shape; these inputs should be rejected consistently as invalid populations.

## Validation
- Not run locally: the execution environment could not clone GitHub (`Could not resolve host: github.com`).
- Connector comparison shows the branch is 4 commits ahead of `main`, 0 behind, touching only the backend support patch and its focused tests.


FlorianPfaff:
Closing this for now because it changes established random.choice empty-population semantics rather than just fixing a small implementation gap. The PR-local JAX failure is tests/backend/test_jax_random_choice_empty_population.py::test_zero_sized_choice_accepts_nonpositive_integer_population, which currently expects this zero-sized draw to be accepted; the branch intentionally raises ValueError instead. That needs a deliberate API decision before it should stay in the open PR queue.

PR #3917, opened 2026-07-06
Add DP birth multi-Bernoulli tracker

FlorianPfaff:
## Summary

Adds a PyRecEst-native Dirichlet-process birth augmentation for the existing multi-Bernoulli tracker.

Implemented pieces:

- `DirichletProcessBirthAtom`, a finite active approximation to a DP birth atom in state space
- `DirichletProcessBirthMultiBernoulliTracker`, a subclass of `MultiBernoulliTracker` that keeps the existing RFS-style labeled Bernoulli recursion and replaces measurement-driven birth creation with a DP/Chinese-restaurant-style birth decision
- `DPBirthMultiBernoulliTracker` and `DPBirthAtom` aliases for shorter imports from the module
- focused regression tests for new-atom creation, clutter suppression, atom reuse, and alias consistency

The implementation is intentionally an LMB/multi-Bernoulli approximation, not a closed-form GLMB/PMBM recursion. It is designed as a concrete integration point for the DP/RFS hybrid idea inside PyRecEst's existing multitarget-tracking code rather than as a standalone package copy.

## Model behavior

For an unassigned measurement, the tracker compares:

- existing birth atom scores: `n_j p(z | atom_j)`
- new birth atom score: `alpha p(z | H_B)`
- clutter intensity through a birth-odds threshold

If birth odds are high enough, the measurement either reuses the best active DP birth atom or creates a new atom, then spawns a Bernoulli component using the selected atom state.

## Testing

- Syntax-compiled the added implementation and test files locally.
- Added `tests/filters/test_dirichlet_process_birth_tracker.py` for repository CI.

I could not run the full PyRecEst test suite in this sandbox because the execution environment cannot clone/install GitHub dependencies from the network.

### DIFF
--- a/src/pyrecest/filters/dirichlet_process_birth_tracker.py
+++ b/src/pyrecest/filters/dirichlet_process_birth_tracker.py
@@ -1,0 +1,407 @@
+"""Dirichlet-process birth augmentation for multi-Bernoulli tracking."""
+
+from __future__ import annotations
+
+import copy
+from dataclasses import dataclass
+from typing import Any
+
+import numpy as np
+
+from pyrecest.distributions import GaussianDistribution
+
+from .multi_bernoulli_tracker import BernoulliComponent, MultiBernoulliTracker
+
+
+@dataclass
+class DirichletProcessBirthAtom:
+    """Finite active approximation to one DP birth atom.
+
+    The atom lives in state space. Its count is the Chinese-restaurant sufficient
+    statistic used by :class:`DirichletProcessBirthMultiBernoulliTracker` when an
+    unassigned measurement can either reuse an existing birth region or instantiate
+    a new region.
+    """
+
+    mean: Any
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Development direction update: the standalone `FlorianPfaff/DP-RFS-Hybrid` repository should be treated as the main research sandbox, with PyRecEst integration proceeding gradually through small, stable upstream PRs. This PR is therefore best viewed as an early integration sketch rather than the primary development branch.

FlorianPfaff:
Follow-up CI fix: the added DP birth tracker tests now skip under non-NumPy backends because the implementation inherits `MultiBernoulliTracker`, which is explicitly NumPy-only. The new head is `b3bc3bdcc285f6550bc167d568a748d2e8a7d288`; CI has restarted.

PR #3932, opened 2026-07-06
Fix LinearDiracDistribution set_mean copy contract

FlorianPfaff:
## Summary
- make `LinearDiracDistribution.set_mean(...)` return a shifted copy instead of mutating the source distribution in place
- validate the requested mean shape consistently with the Gaussian mixture setter contract
- add regression tests that verify the original Dirac distribution remains unchanged and invalid mean dimensions are rejected

## Bug fixed
`LinearDiracDistribution.set_mean(...)` previously returned `None` and modified `self.d` in place. That was inconsistent with the value-returning `set_mean` API used by Gaussian distributions and mixtures, and made `shifted = dirac.set_mean(...)` unusable.

## Tests
Not run locally; changes were made through the GitHub connector.


FlorianPfaff:
Closing this one for now. The branch changes LinearDiracDistribution.set_mean from the established in-place contract into a copy-returning API, but the existing distribution/conversion tests still rely on the in-place behavior and fail across the matrix. That needs a deliberate API design decision and coordinated test/caller updates rather than a quick cleanup fix.

PR #3953, opened 2026-07-06
Fix PyTorch dot NumPy contraction contract

FlorianPfaff:
## Summary
- Fix the PyTorch backend `dot` runtime contract so matrix/matrix, vector/matrix, matrix/vector, and batched contractions follow NumPy `dot` semantics.
- Override the previous row-wise inner-product patch with a tensor contraction over the last axis of the left operand and the second-to-last axis of the right operand.
- Add a PyTorch-backend regression test covering matrix/matrix, vector/matrix, and batched matrix cases.

## Bug
The existing PyTorch linear-helper patch used `einsum("...i,...i->...", a, b)` for higher-rank operands. For two matrices this returns row-wise inner products such as `[17, 53]` for `[[1, 2], [3, 4]] dot [[5, 6], [7, 8]]`, while NumPy returns the matrix product `[[19, 22], [43, 50]]`.

## Testing
- Added `tests/backend_support/test_pytorch_dot_numpy_contract.py`.
- I could not run the repository test suite in this environment because the repository is only accessible through the GitHub connector, not as a local checkout.

### DIFF
--- a/src/pyrecest/_backend_runtime_patches.py
+++ b/src/pyrecest/_backend_runtime_patches.py
@@ -108,3 +108,63 @@ def patch_pytorch_repeat_numpy_contract() -> None:
     raw_pytorch.repeat = repeat
     if active_pytorch_backend:
         backend.repeat = repeat
+
+
+def patch_pytorch_dot_numpy_contract() -> None:
+    """Make PyTorch ``dot`` follow NumPy's matrix/vector contraction contract."""
+
+    try:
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this as superseded by #3977, which covers the same PyTorch dot NumPy semantics and extends the fix to outer semantics with newer backend-support tests. This older branch now conflicts in runtime patch plumbing, so keeping the newer PR is the lower-risk path.

PR #3959, opened 2026-07-06
Stabilize NumPy multinomial probability normalization

FlorianPfaff:
## Summary
- Normalize NumPy multinomial probability weights by their maximum before summing.
- Preserve support for finite, unnormalized probability vectors whose direct float64 sum would overflow.
- Add a NumPy-specific regression test using large finite weights.

## Testing
- Not run locally; changes were made through the GitHub connector after syntax-level review of the edited snippets.

### DIFF
--- a/src/pyrecest/_backend/numpy/random.py
+++ b/src/pyrecest/_backend/numpy/random.py
@@ -80,6 +80,21 @@ def _validate_multinomial_sample_count(n):
     return count
 
 
+def _normalize_probability_array(probability_array):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #4036. This PR handles the large finite NumPy `multinomial` probability overflow case; #4036 is the newer branch on current `main` and covers the same `multinomial` normalization plus the matching `choice` path with shared helper coverage, so it is the better branch to keep.

PR #3960, opened 2026-07-06
Fix PyTorch take axis validation

FlorianPfaff:
## Summary
- Add a PyTorch backend compatibility patch for stricter `take` axis handling.
- Wire the patch into the existing stability setup.

## Validation
- Branch is 2 commits ahead of current `main` and 0 behind.
- Existing focused regression coverage is in `tests/backend_support/test_pytorch_take_axis_contract.py`.
- Full local pytest was not run in this connector environment; CI should run the regression tests.


FlorianPfaff:
Closing this as superseded by #4004. #4004 carries the same PyTorch `take` axis contract fix through the current runtime patch hook and preserves the stricter boolean/float-axis behavior, so keeping both open would duplicate the same backend contract change.

PR #3980, opened 2026-07-07
Fix shared NumPy squeeze non-unit axis contract

FlorianPfaff:
## Summary
- Fix shared NumPy/autograd `squeeze(..., axis=...)` runtime patching so explicitly requested axes with size other than one raise `ValueError` instead of silently returning the input.
- Synchronize the patched implementation into the active raw backend module (`pyrecest._backend.numpy` or `pyrecest._backend.autograd`) so raw-backend calls obey the same contract as the public backend.
- Keep existing scalar-array axis, tuple-axis, duplicate-axis, and out-of-bounds validation behavior intact.
- Add public-backend and raw-NumPy regression coverage for non-singleton axes.

## Bug
`backend.squeeze(array([[1], [2]]), axis=0)` previously returned the original `(2, 1)` array. NumPy raises `ValueError` for an explicitly requested non-singleton axis, and the PyRecEst NumPy/autograd facade should preserve that contract. The raw NumPy backend had the same silent no-op behavior, so this patch also keeps `pyrecest._backend.numpy.squeeze(...)` synchronized with the fixed runtime behavior.

## Tests
- Not run locally in this environment because the repository cannot be cloned here; added focused pytest regression coverage in `tests/test_shared_numpy_squeeze.py`.

### DIFF
--- a/src/pyrecest/_backend_runtime_patches.py
+++ b/src/pyrecest/_backend_runtime_patches.py
@@ -2,6 +2,102 @@
 
 from __future__ import annotations
 
+from operator import index as _operator_index
+
+
+def patch_shared_numpy_squeeze_axis_contract() -> None:
+    """Make shared NumPy/autograd squeeze reject non-unit requested axes."""
+
+    try:
+        import pyrecest.backend as backend  # pylint: disable=import-outside-toplevel
+        import pyrecest._backend._shared_numpy as shared_numpy  # pylint: disable=import-outside-toplevel
+    except ModuleNotFoundError:  # pragma: no cover - source tree corruption only
+        return
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this for now. The branch is merge-conflicted and its failures are PR-local rather than the shared PyTorch argsort issue: NumPy fails tests/test_backend_arraylike_contracts.py::test_squeeze_non_singleton_axis_is_noop because the PR intentionally flips that established contract, and JAX collection fails with a circular import / missing pyrecest.backend cascade after the runtime patch. This needs a deliberate backend-contract/API decision and a cleaner implementation before it should remain in the active PR queue.

PR #3981, opened 2026-07-07
Fix PyTorch concatenate axis=None contract

FlorianPfaff:
## Summary
- patch the PyTorch backend runtime contract so `concatenate(..., axis=None)` flattens each input before concatenation, matching NumPy/JAX semantics
- install the patch through the existing evidence/runtime backend patch path
- add raw and public PyTorch regression coverage for `axis=None`

## Tests
- Not run locally; added `tests/backend_support/test_pytorch_concatenate_contract.py` for CI coverage.


FlorianPfaff:
Closing this as superseded by #4031, which now covers the PyTorch concatenate axis=None behavior as part of the broader stack/concatenate helper contract and already has the relevant regression coverage. Keeping both open would leave duplicate conflicting runtime-patch changes.

PR #3991, opened 2026-07-07
Fix PyTorch squeeze non-singleton axis contract

FlorianPfaff:
## Summary
- Add a PyTorch backend runtime patch so `squeeze(..., axis=...)` rejects explicitly requested non-singleton axes instead of silently returning the input.
- Wire the patch into the existing late backend-runtime patch hook used from `evidence.py`.
- Add public-backend and raw-PyTorch regression coverage for non-singleton axis handling.

## Bug
The raw PyTorch backend currently delegates `backend.squeeze(x, axis=...)` to `torch.squeeze(x, dim=...)`. PyTorch silently leaves tensors unchanged when the requested dimension has size other than one, but NumPy-compatible `squeeze(axis=...)` raises `ValueError` for an explicitly requested non-singleton axis. This made the PyTorch backend diverge from the backend contract.

## Tests
- Not run locally in this environment because the repository cannot be cloned from GitHub here; added focused pytest coverage in `tests/backend_support/test_pytorch_squeeze_axis_contract.py`.

### DIFF
--- a/src/pyrecest/_backend_runtime_patches.py
+++ b/src/pyrecest/_backend_runtime_patches.py
@@ -236,6 +236,89 @@ def patch_pytorch_edge_pad_contract() -> None:
         backend.pad = pad
 
 
+def _pytorch_squeeze_axes(axis, ndim, numpy_module, torch_module):
+    """Return normalized NumPy-style squeeze axes for a PyTorch tensor."""
+
+    if torch_module.is_tensor(axis):
+        axis = axis.detach().cpu().numpy()
+
+    axis_array = numpy_module.asarray(axis)
+    if axis_array.shape == ():
+        try:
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this because the patch changes the established PyRecEst backend contract for squeeze. The existing contract test  expects an explicitly requested non-singleton axis to leave the input unchanged, and the PyTorch failures are exactly that contract conflict rather than an incidental CI issue. Changing squeeze semantics would need a deliberate cross-backend contract update, not a narrow PyTorch-only cleanup PR.

FlorianPfaff:
Correction to the close note: the relevant existing contract test is `test_squeeze_non_singleton_axis_is_noop`, which expects an explicitly requested non-singleton squeeze axis to leave the input unchanged.

PR #3995, opened 2026-07-07
Reject infinite max time deltas

FlorianPfaff:
## Summary
- reject infinite `max_time_delta_s` values in calibration validation
- keep `None` as the explicit no-limit sentinel
- add regression coverage for interpolation and bias-training helpers

## Testing
- Not run locally; patch was applied via the GitHub connector.


FlorianPfaff:
Closing this rather than patching it because it reverses the current calibration API contract: existing coverage in tests/calibration/test_time_offset_bias.py explicitly treats np.inf max_time_delta_s as an unbounded value. The branch only updates the narrower validation test and leaves those broader behavior tests failing, so this looks like a noisy/low-value semantics change rather than an easy valuable fix.

PR #4001, opened 2026-07-07
Fix PyTorch stack helpers on empty sequences

FlorianPfaff:
## Summary
- Add an explicit NumPy-compatible empty-input guard for PyTorch `hstack`, `vstack`, `column_stack`, and `dstack` compatibility helpers.
- Preserve the existing device-normalization behavior for non-empty stack-helper calls.
- Add backend-portable subprocess regression coverage for public PyTorch and raw PyTorch helpers under NumPy backend selection.

## Bug fixed
The PyTorch stack-helper compatibility hook normalized inputs and then delegated empty sequences to `torch.cat([])`. That produces a PyTorch runtime error instead of NumPy's stable `ValueError: need at least one array to concatenate` contract for `np.hstack([])`, `np.vstack([])`, `np.column_stack([])`, and `np.dstack([])`.

## Validation
- `python -m py_compile /tmp/_pytorch_stack_helpers_device_contract.py /tmp/test_pytorch_stack_helpers_empty_sequence_contract.py`
- Isolated stack-helper smoke check using a minimal raw PyTorch stub: verified all four helpers raise the intended `ValueError` for `[]` and preserve the existing non-empty examples.
- Full repository pytest was not run locally because this environment cannot clone GitHub directly; changes were applied through the GitHub connector.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_stack_helpers_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_stack_helpers_device_contract.py
@@ -6,6 +6,7 @@ import importlib
 
 
 _STACK_HELPER_NAMES = ("hstack", "vstack", "column_stack", "dstack")
+_EMPTY_STACK_HELPER_ERROR = "need at least one array to concatenate"
 
 
 def _raw_pytorch_module():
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #4031. #4031 covers the same PyTorch stack-helper empty-sequence contract and expands the fix to concatenate/stack with consolidated regression coverage, so keeping this older dirty branch open would duplicate the same backend-support change.

PR #4005, opened 2026-07-07
Reject text scalar ROI unmatched values

FlorianPfaff:
## Summary
- Reject text/bytes scalar `unmatched_value` inputs in the ROI assignment validation shim instead of coercing numeric-looking strings through `float()`.
- Extend the ROI assignment regression test to cover Python and NumPy text scalar inputs that previously slipped through when they did not collide with a real column index.

## Bug fixed
`assign_by_similarity_matrix(..., unmatched_value=...)` is documented and validated as requiring an integer sentinel outside the valid column range. The validator rejected booleans, non-integral floats, NaN, and non-scalar arrays, but text scalars such as `"2"` or `b"-1"` could still be converted by `float()` and accepted. This could silently turn malformed configuration/text input into a valid sentinel.

## Validation
- Added regression coverage in `tests/test_roi_assignment_unmatched_value.py`.
- Full test suite not run in this connector-only environment.

### DIFF
--- a/src/pyrecest/utils/__init__.py
+++ b/src/pyrecest/utils/__init__.py
@@ -41,11 +41,15 @@ if not hasattr(_assignment_module, "_solve_subproblem_without_full_assignment"):
 
 def _normalize_roi_unmatched_value(unmatched_value):
     value_array = _roi_assignment_module.asarray(unmatched_value)
-    if value_array.shape != ():
+    if (
+        value_array.shape != ()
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #4049. Both PRs reject text/bytes ROI `unmatched_value` sentinels; #4049 is the newer branch on current `main` with the same functional fix and focused regression coverage, so keeping both open only duplicates CI and conflict surface.

PR #4034, opened 2026-07-07
Stabilize NumPy probability normalization

FlorianPfaff:
## Summary
- Normalize NumPy `choice` and `multinomial` probability weights using max-scaling before summing.
- Share the stable probability-normalization helper between shared NumPy `choice` and NumPy `multinomial`.
- Add regression coverage for large finite unnormalized weights that used to overflow during validation.

## Bug fixed
The NumPy backend converted probability weights to float and summed them directly. Large but finite unnormalized weights such as `[1e308, 1e308]` therefore overflowed to `inf` and were rejected as invalid, even though they can be safely normalized by scaling first. This mirrors the stable normalization pattern already used in the JAX backend.

## Validation
- `python -m py_compile` on the patched source files and new regression test.
- Branch comparison shows 3 changed files.

### DIFF
--- a/src/pyrecest/_backend/_shared_numpy/random.py
+++ b/src/pyrecest/_backend/_shared_numpy/random.py
@@ -202,6 +202,19 @@ def _integer_choice_population_size(a_array):
     return None
 
 
+def _normalize_probability_values(values):
+    if _np.any(values < 0.0) or _np.any(~_np.isfinite(values)):
+        raise ValueError("probabilities do not sum to a positive value")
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing this as superseded by #4036. Both PRs implement the same stable NumPy probability-normalization fix for large finite weights; #4036 is the newer branch and keeps the regression coverage in the existing NumPy random backend test file, so keeping one version reduces duplicate CI and merge conflict noise.

PR #4067, opened 2026-07-08
Fix PyTorch dot matrix product contract

FlorianPfaff:
## Summary
- make the PyTorch dot/outer device compatibility hook dispatch all scalar/vector/matrix pairs through `torch.matmul` after scalar handling
- add regression coverage showing raw and public PyTorch `dot` preserve the preferred non-CPU device and return the NumPy-style matrix product shape/value for 2-D inputs

## Bug fixed
The active PyTorch compatibility hook still treated two matrix operands as a batched inner product via `einsum("...i,...i->...")`. For 2-D inputs this returned a row-wise contraction or failed on valid matrix-product shapes instead of following NumPy `dot(A, B)` semantics. The patch uses `torch.matmul` for all non-scalar operands where both inputs have at most two dimensions, preserving vector-vector, matrix-vector, vector-matrix, and matrix-matrix behavior.

## Validation
- Syntax-compiled the modified hook and regression test locally with `py_compile`.
- Compared branch against `main`: 2 commits ahead, 1 behind; changed files are limited to the PyTorch dot/outer compatibility hook and its focused backend-support test.
- Full repository tests were not run locally because this environment cannot resolve `github.com`; CI should run the focused PyTorch backend-support regression.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_dot_outer_device_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_dot_outer_device_contract.py
@@ -118,8 +118,8 @@ def patch_pytorch_dot_outer_device_contract() -> None:
         a, b = _promoted_pair(raw_pytorch, torch, a, b)
         if a.ndim == 0 or b.ndim == 0:
             return torch.multiply(a, b)
-        if a.ndim == 1 and b.ndim == 1:
-            return torch.dot(a, b)
+        if a.ndim <= 2 and b.ndim <= 2:
+            return torch.matmul(a, b)
         if b.ndim == 1:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this one because the patch changes PyTorch dot semantics from PyRecEst's established last-axis inner-product contract to matrix multiplication. The failing checks are not incidental: existing backend contract tests expect 2-D dot inputs to reduce over the last axis, while matrix multiplication is covered by backend.matmul. Keeping this open would conflict with the current public backend contract and overlaps with the separate PyTorch dot/outer cleanup work.

PR #4071, opened 2026-07-08
Fix Pareto front eligibility for nonnumeric objectives

FlorianPfaff:
## Summary
- Treat nonnumeric objective values as non-comparable when deciding whether a row is eligible for the Pareto front.
- Reuse the same `_lookup_numeric(...)` coercion path for front eligibility that dominance comparisons already use.
- Add a regression test covering rows whose objective values are text/bytes rather than numeric or NaN.

## Bug fixed
With `allow_missing=True`, `_has_front_eligible_objectives(...)` checked raw values with `record.get(...)` and `pd.isna(...)`. Text or bytes objective values are not missing under that raw check, so a row with no numeric comparable objectives could be marked front-eligible. Later, `record_dominates(...)` coerces those same values through `_lookup_numeric(...)` and treats them as missing/non-comparable, meaning no other row could dominate the malformed row. The result was a spurious Pareto-front entry.

The patch makes front eligibility use `_lookup_numeric(...)` as well, so rows need at least one numeric comparable objective before they can appear on the front.

## Validation
- `python -m py_compile /tmp/pareto.py /tmp/test_pareto_missing_front_eligibility.py`
- In a minimal local package harness: `PYTHONPATH=/tmp/pyrecest_testpkg python -m pytest -q /tmp/test_pareto_missing_front_eligibility.py` → 3 passed
- Confirmed the new regression fails against the previous raw-value eligibility logic: 1 failed, 2 passed
- GitHub compare: branch is 2 commits ahead of `main`, 0 behind; changed files are limited to the Pareto utility and focused regression test.

### DIFF
--- a/src/pyrecest/evaluation/pareto.py
+++ b/src/pyrecest/evaluation/pareto.py
@@ -301,7 +301,7 @@ def _has_front_eligible_objectives(
 ) -> bool:
     if allow_missing:
         return any(
-            not _is_missing(record.get(objective, np.nan))
+            not _is_missing(_lookup_numeric(record, objective))
             for objective in objectives
         )
     return all(
--- a/tests/evaluation/test_pareto_missing_front_eligibility.py
+++ b/tests/evaluation/test_pareto_missing_front_eligibility.py
@@ -28,6 +28,29 @@ def test_pareto_front_excludes_rows_without_comparable_objectives() -> None:
     assert mask.tolist() == [True, False]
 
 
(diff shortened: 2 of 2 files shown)


FlorianPfaff:
Closing this for now because it changes the existing allow_missing Pareto semantics. The CI failure is PR-local: tests/evaluation/test_pareto_objective_scalar_validation.py still expects a row with a nonnumeric/missing objective to remain eligible, while this branch excludes it. That is more of a policy/API semantics decision than an easy cleanup fix.

PR #4074, opened 2026-07-08
Reject datetime-like time offset inputs

FlorianPfaff:
## Summary
- reject NumPy `datetime64`/`timedelta64` scalar values in time-offset scalar validators before float coercion
- reject object arrays containing datetime-like values in time-series inputs
- reject complex object scalars explicitly in scalar validators, preventing NumPy from discarding imaginary parts during `float(...)` conversion
- add focused regressions for object-array times, scalar offsets, and max-time-delta inputs

## Bug fixed
`time_offset.py` previously allowed some non-real time-like values through float coercion. In particular, nanosecond `np.timedelta64` values can expose integer-like payloads, so inputs such as `np.timedelta64(1, "ns")` could be interpreted as `1.0` seconds for `offset_s` or `max_time_delta_s`. Object arrays containing `np.datetime64`/`np.timedelta64` could also be converted to floats by `np.asarray(..., dtype=float)`. The patch rejects these datetime-like inputs before conversion.

The same scalar path now also rejects complex object scalars explicitly so NumPy cannot silently discard the imaginary part under a `ComplexWarning`.

## Validation
- `python -m py_compile /mnt/data/time_offset_patched.py`
- isolated focused unittest mirror: `PYTHONPATH=/mnt/data python -m unittest /mnt/data/test_time_offset_real_numeric_validation_patched.py -v` → 5 passed

Full in-repository pytest was not run locally because this environment cannot resolve `github.com` for a checkout; CI should run the in-tree regression.

### DIFF
--- a/src/pyrecest/calibration/time_offset.py
+++ b/src/pyrecest/calibration/time_offset.py
@@ -10,6 +10,7 @@ import numpy as np
 
 _ERROR_METRIC_NAMES = frozenset({"max", "mean", "p95", "rmse", "std"})
 _ERROR_METRIC_MESSAGE = "metric must be one of 'max', 'mean', 'p95', 'rmse', or 'std'"
+_TEMPORAL_SCALAR_TYPES = (np.datetime64, np.timedelta64)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this in favor of #4075, which applies the same temporal-input guard through a shared detector and adds broader coverage for offset scalars, object-wrapped temporal times, max-time-delta values, reference values, and aggregate summaries. The extra complex-object scalar checks here do not appear to add distinct behavior beyond the existing float-conversion rejection path, so keeping both PRs open would likely duplicate work and create conflict noise.

PR #4077, opened 2026-07-08
Reject temporal scalars in time-offset calibration

FlorianPfaff:
## Summary
- reject NumPy `datetime64`/`timedelta64` scalars before time-offset scalar coercion
- prevent object arrays containing temporal scalars from being treated as numeric times or sweep offsets
- add focused regression coverage for `offset_s`, `max_time_delta_s`, `times_s`, and aggregate sweep offsets

## Bug fixed
NumPy `datetime64`/`timedelta64` values with nanosecond units can expose integer-like scalar values and `float(...)` can turn them into epoch/duration numbers. `time_offset` validators accepted some direct scalar and object-array temporal inputs, so malformed timestamps or durations could be silently interpreted as seconds.

## Validation
- `python -m py_compile /tmp/time_offset.py`
- Focused local unittest mirror: `python -m unittest /tmp/test_time_offset_temporal.py -v` → 4 tests passed
- GitHub compare: branch is 2 commits ahead of `main`, 0 behind; changed only `src/pyrecest/calibration/time_offset.py` and `tests/calibration/test_time_offset_temporal_validation.py`

Full repository pytest was not run locally because this sandbox cannot resolve `github.com`; CI should run the in-tree regression.

### DIFF
--- a/src/pyrecest/calibration/time_offset.py
+++ b/src/pyrecest/calibration/time_offset.py
@@ -10,6 +10,18 @@ import numpy as np
 
 _ERROR_METRIC_NAMES = frozenset({"max", "mean", "p95", "rmse", "std"})
 _ERROR_METRIC_MESSAGE = "metric must be one of 'max', 'mean', 'p95', 'rmse', or 'std'"
+_REJECTED_SCALAR_VALUE_TYPES = (
+    type(None),
+    bool,
+    np.bool_,
+    str,
+    bytes,
+    bytearray,
+    complex,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this in favor of #4075, which now carries the maintained time-offset temporal-input fix with broader coverage for offset scalars, object-wrapped times, max-time-delta values, reference values, and aggregate summaries. Keeping both branches open would duplicate the same area and create avoidable conflict noise.

PR #4086, opened 2026-07-08
Reject temporal Pareto scalar inputs

FlorianPfaff:
## Summary
- reject NumPy `datetime64`/`timedelta64` scalar controls before Pareto `eps` and constraint-threshold coercion
- treat object-wrapped temporal objective scalars as nonnumeric/missing instead of comparable floats
- add focused regression coverage for temporal `eps`, constraint thresholds, and object temporal objectives

## Bug fixed
`np.timedelta64` values can expose integer nanosecond payloads through `.item()` / `float(...)`. That meant malformed temporal values such as `np.timedelta64(1, "ns")` could be silently accepted as Pareto tolerances or constraint thresholds, and object-wrapped temporal objective values could be compared numerically.

## Validation
- `python -m py_compile /mnt/data/pareto.py /mnt/data/test_pareto_temporal_scalar_validation.py`
- focused local smoke of the patched Pareto helper paths for temporal `eps`, temporal thresholds, object temporal objectives, and an existing min/max Pareto-front example

Full in-repository pytest was not run locally because this sandbox cannot clone from `github.com`; CI should run the added in-tree regression.

### DIFF
--- a/src/pyrecest/evaluation/pareto.py
+++ b/src/pyrecest/evaluation/pareto.py
@@ -24,6 +24,10 @@ def _is_text_scalar(value: Any) -> bool:
     return isinstance(value, (str, bytes, np.str_, np.bytes_))
 
 
+def _is_temporal_scalar(value: Any) -> bool:
+    return isinstance(value, (np.datetime64, np.timedelta64))
+
+
 def pareto_front_indices(
     table: pd.DataFrame,
     objectives: Sequence[str],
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this as superseded by #4085. Both PRs target the same Pareto temporal scalar gap, and #4085 keeps the scalar coverage while also handling datetime64 inputs and temporal-dtyped constraint columns, so keeping both open would duplicate CI and review work.

PR #4127, opened 2026-07-09
Reject invalid PyTorch linalg tolerances

FlorianPfaff:
## Summary
- add a runtime patch around the PyTorch linalg `pinv` and `matrix_rank` tolerance arguments
- reject boolean and NumPy temporal tolerances before scalar unwrapping can turn them into numeric thresholds
- add focused regression coverage for `rcond`/`atol`/`rtol` on `pinv` and `tol`/`atol`/`rtol` on `matrix_rank`

## Bug fixed
The PyTorch linalg bridge normalizes NumPy scalar and scalar-array tolerances with `.item()`. For nanosecond-resolution `np.datetime64` / `np.timedelta64`, that can expose the raw integer payload, and boolean scalars can be accepted by PyTorch as numeric thresholds. As a result, invalid tolerances such as `np.timedelta64(1, "ns")`, `np.datetime64("1970-01-01T00:00:00.000000001")`, or `True` could silently become an effective tolerance of `1`, causing `pinv`/`matrix_rank` to zero out small singular values or rank unexpectedly instead of rejecting malformed input.

## Validation
- local `py_compile` for the new patch file, evidence wiring, and regression test
- isolated local smoke test of the wrapper rejection logic for Python/NumPy/Torch boolean tolerances and NumPy temporal tolerances
- GitHub compare: branch is 3 commits ahead of `main`, 0 behind, changing only the tolerance patch, evidence hook, and focused regression test

Full in-repository pytest was not run locally because this environment cannot clone GitHub; CI should run the new test in-repo.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_linalg_tolerance_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_linalg_tolerance_contract.py
@@ -1,0 +1,128 @@
+"""Runtime patch for PyTorch linalg tolerance validation."""
+
+from __future__ import annotations
+
+from functools import wraps
+
+import numpy as np
+
+_TOLERANCE_TYPE_MESSAGE = (
+    "linalg tolerances must be real numeric, not boolean or temporal"
+)
+_BOOLEAN_TYPES = (bool, np.bool_)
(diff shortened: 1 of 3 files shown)


FlorianPfaff:

PR #4128, opened 2026-07-09
Reject temporal PyTorch linalg tolerances

FlorianPfaff:
## Summary
- add a PyTorch linalg tolerance contract patch that rejects NumPy `datetime64`/`timedelta64` tolerance inputs before scalar unwrapping can expose raw integer payloads
- wire the patch into the existing backend patch bootstrap in `pyrecest.evidence`
- add regression coverage for `matrix_rank` and `pinv` temporal tolerances while preserving existing NumPy scalar/vector numeric tolerance support

## Validation
- Not run locally: the sandbox could not clone GitHub (`Could not resolve host: github.com`), so this was applied through the GitHub connector only.


FlorianPfaff:

PR #4139, opened 2026-07-10
Preserve caller RNG state during scenario generation

FlorianPfaff:
## Summary
- snapshot the selected backend RNG state and NumPy RNG state before scenario generation
- restore both states in a `finally` block so successful and failed generation cannot perturb caller-owned random streams
- add focused regressions covering normal generation and exceptions

## Bug fixed
`generate_simulated_scenarios()` seeds both `pyrecest.backend.random` and `numpy.random` for every run, but previously left those global RNGs at the state reached by the last generated scenario. Calling the helper therefore changed subsequent random values in unrelated application code. The behavior was also observable when generation raised midway through a run.

The generated scenarios remain deterministic for the configured per-run seeds; only the unintended process-wide side effect is removed.

## Validation
- branch is 2 commits ahead of current `main` and 0 behind
- Python syntax compilation passed for the modified module and new regression test
- direct NumPy-backend restoration probe passed for success and exception paths
- direct PyTorch/NumPy restoration probe passed for success and exception paths

Full repository tests are delegated to GitHub Actions because the execution environment cannot clone GitHub.

### DIFF
--- a/src/pyrecest/evaluation/generate_simulated_scenarios.py
+++ b/src/pyrecest/evaluation/generate_simulated_scenarios.py
@@ -14,6 +14,17 @@ def _seed_simulation_rngs(seed):
     np.random.seed(seed)
 
 
+def _capture_simulation_rng_states():
+    """Capture caller-owned backend and NumPy RNG states."""
+    return random.get_state(), np.random.get_state()
+
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #4140, opened 2026-07-10
Preserve caller RNG state during simulation generation

FlorianPfaff:
## Summary

- save and restore both `pyrecest.backend.random` and NumPy RNG state around `generate_simulated_scenarios`
- restore the states through a `finally` block, including when ground-truth or measurement generation raises
- add regression tests for successful and failing generation paths

## Bug

`generate_simulated_scenarios` reseeds process-wide RNGs for every requested simulation seed. Before this change, the final simulation seed and all random draws made during generation leaked into the caller, so unrelated code produced different random sequences after generating scenarios. Exceptions also left both RNGs contaminated.

## Compatibility

Simulation outputs for a given `all_seeds` configuration remain deterministic and unchanged; only the external RNG side effect is removed.

### DIFF
--- a/src/pyrecest/evaluation/generate_simulated_scenarios.py
+++ b/src/pyrecest/evaluation/generate_simulated_scenarios.py
@@ -18,7 +18,7 @@ def generate_simulated_scenarios(
     simulation_params,
 ):
     """
-    Generate simulated scenarios.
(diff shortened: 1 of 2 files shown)


FlorianPfaff:

PR #4206, opened 2026-07-11
Stabilize large-kappa Gauss von Mises formulas

FlorianPfaff:
## Summary
- evaluate the Gauss–von Mises circular density with an exponent shift and the scaled modified Bessel function `ive`
- use scaled Bessel ratios in the Horwood deterministic sigma-point construction
- add regressions for `kappa=1000` covering finite mode density, finite sigma points/weights, and normalized weights

## Bug fixed
`GaussVonMisesDistribution` accepted large finite concentrations, but two formulas still used unscaled `iv` values. For `kappa=1000`, both `exp(kappa)` and `iv(0, kappa)` overflow in the density, producing `inf / inf -> NaN`. The Horwood `I_p(kappa) / I_0(kappa)` ratios had the same problem.

For nonnegative `kappa`, rewriting

`exp(kappa*cos(delta)) / I_0(kappa)`

as

`exp(kappa*(cos(delta)-1)) / ive(0, kappa)`

is algebraically equivalent while remaining finite. Ratios of `ive` values are likewise identical to ratios of `iv` values because the common exponential scaling cancels.

## Validation
- reconstructed the current source file and verified its Git blob SHA matched `70696cee334ac057644a1bebc22294682a62ab9f` before editing
- `python -m py_compile` on the modified module and new regression file
- standalone numerical comparison at moderate concentrations matched the original expression to floating-point precision
- at `kappa=1000`, the stabilized mode density is finite (`12.61408496162745` for the circular factor), and the Horwood weights are finite and sum to one

The full repository test suite was not run locally because this environment has no PyRecEst checkout; GitHub CI should exercise the in-repository regressions.

### DIFF
--- a/src/pyrecest/distributions/cart_prod/gauss_von_mises_distribution.py
+++ b/src/pyrecest/distributions/cart_prod/gauss_von_mises_distribution.py
@@ -31,7 +31,7 @@ from pyrecest.backend import (
     zeros_like,
 )
 from scipy.integrate import nquad
-from scipy.special import iv
+from scipy.special import ive
 from scipy.stats import vonmises as _vonmises
 
 from ..nonperiodic.gaussian_distribution import GaussianDistribution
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
The intended fix is already present on current `main`: `GaussVonMisesDistribution` uses scaled `ive` in both the PDF and Horwood ratios, and `tests/distributions/test_gauss_von_mises_large_kappa.py` covers finite large-κ densities, sigma points, and normalized weights. Closing this stale conflicted PR as already incorporated.

PR #4208, opened 2026-07-12
Stabilize toroidal von Mises sine densities at large concentration

FlorianPfaff:
## Summary

- evaluate the bivariate von Mises sine normalizer with exponentially scaled modified Bessel functions
- retain the common `exp(kappa[0] + kappa[1])` factor in log form
- evaluate the PDF with both concentration exponents shifted by `-kappa`
- add a focused regression for the independent `kappa = [1000, 1000]` case

## Bug

`ToroidalVonMisesSineDistribution` computed its normalization series from unscaled `scipy.special.iv` values and evaluated the density as `C * exp(raw_exponent)`.

For large but valid concentrations, `I_0(kappa)` and the density exponential overflow separately. At `kappa = [1000, 1000]` with `lambda_ = 0`, construction encounters a non-finite normalization term; equivalently, the old density formulation would combine an underflowed reciprocal normalizer with `exp(2000)` at the mode.

## Fix

Use `scipy.special.ive` in the normalization series. Every product of the two Bessel factors shares the scale `exp(kappa[0] + kappa[1])`, so the implementation factors that scale out and exposes the full normalizer through `log_norm_const`.

The PDF is evaluated as

```python
exp(
    kappa0 * (cos(delta0) - 1)
    + kappa1 * (cos(delta1) - 1)
    + coupling
    - log(scaled_norm_const)
)
```

which is algebraically identical while keeping the independent high-concentration mode finite.

`norm_const` continues to expose the ordinary-space value and returns `inf` when the mathematically valid normalizer exceeds floating-point range; PDF evaluation no longer depends on that overflowing value.

## Impact

Highly concentrated bivariate von Mises sine distributions can now be constructed and evaluated without avoidable `inf`/`NaN` intermediates. Moderate-concentration values and the zero-concentration limiting formula remain unchanged to floating-point precision.

## Validation

- exact changed module syntax-compiled successfully
- focused exact-module smoke test passed for the `kappa = [1000, 1000]`, `lambda_ = 0` regression
- mode density matched the scaled-Bessel reference: `159.1151394191558`
- moderate-parameter PDF values matched the previous expression to approximately `4.4e-16` relative error
- zero-concentration normalizer remained finite
- branch is two commits ahead of current `main`, zero behind, and changes only the implementation and its focused test

### DIFF
--- a/src/pyrecest/distributions/hypertorus/toroidal_von_mises_sine_distribution.py
+++ b/src/pyrecest/distributions/hypertorus/toroidal_von_mises_sine_distribution.py
@@ -2,8 +2,8 @@
 # pylint: disable=no-name-in-module,no-member
 import math
 
-from pyrecest.backend import sin
-from scipy.special import gammaln, iv
+from pyrecest.backend import array, cos, exp, sin
+from scipy.special import gammaln, ive
 
 from .abstract_toroidal_bivar_vm_distribution import (
     AbstractToroidalBivarVMDistribution,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4212. The canonical PR covers the same large-concentration sine-model stabilization and extends it to the cosine model and shared log-normalizer path while preserving the existing ordinary-parameter path. Closing the narrower duplicate.

PR #4215, opened 2026-07-12
Reject non-integer wrapped exponential moment orders

FlorianPfaff:
## Summary
- validate `WrappedExponentialDistribution.trigonometric_moment()` orders before evaluating the analytic formula
- reject fractional, boolean, and textual orders consistently with the circular moment API
- preserve support for negative integer orders and NumPy integer scalars

## Bug
`WrappedExponentialDistribution.trigonometric_moment()` accepted any value that happened to work in arithmetic. Inputs such as `0.5` and `True` were silently evaluated, while textual input failed incidentally during division.

Circular trigonometric moments are Fourier coefficients indexed by integers, so accepting non-integer orders is mathematically invalid and inconsistent with the validation used by other circular distributions.

## Fix
Require `n` to be a non-boolean `numbers.Integral`, normalize accepted integer-like values to a Python `int`, and then evaluate the existing closed-form expression unchanged.

## Validation
- regression rejects `0.5`, `True`, and `"1"` with a clear `ValueError`
- regression verifies the `-2` moment is the conjugate of the `+2` moment
- standalone validation confirmed NumPy integer acceptance and the conjugacy identity
- branch is 2 commits ahead of current `main`, 0 behind, and changes only the wrapped exponential implementation and its existing test module

The full repository test matrix is delegated to GitHub Actions because this runtime cannot resolve `github.com` for a local checkout and does not provide the GitHub CLI.

### DIFF
--- a/src/pyrecest/distributions/circle/wrapped_exponential_distribution.py
+++ b/src/pyrecest/distributions/circle/wrapped_exponential_distribution.py
@@ -124,6 +124,9 @@ class WrappedExponentialDistribution(AbstractCircularDistribution):
         return self.lambda_ * exp(-self.lambda_ * xs) * self._normalization_const
 
     def trigonometric_moment(self, n):
+        if isinstance(n, bool) or not isinstance(n, Integral):
+            raise ValueError("n must be an integer.")
+        n = int(n)
         return 1.0 / (1.0 - 1j * n / self.lambda_)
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4218. The canonical PR covers the same integer-order validation and additionally rejects NumPy booleans while explicitly testing NumPy integer acceptance. Closing this duplicate.

PR #4220, opened 2026-07-12
Normalize NumPy scalar axes in shared PyTorch reductions

FlorianPfaff:
## Summary
- patch the shared `_backend._common` reduction-axis normalizer alongside the public loader and raw PyTorch backend
- factor normalizer installation into one idempotent helper
- add subprocess regression coverage for `min(..., axis=(np.asarray(0),))` on a Torch tensor under the NumPy public backend

## Bug
The shared Torch-compatible `min` implementation resolves tuple axes through `pyrecest._backend._common._normalize_reduction_axes`. The existing runtime patch updated the loader export and raw PyTorch module, but not the shared module whose globals are used by `_common.min`.

For a valid NumPy-style tuple such as `(np.asarray(0),)`, the unpatched shared normalizer retained the zero-dimensional ndarray instead of converting it with `operator.index`. It then attempted to construct a set from the normalized axes and raised:

```text
TypeError: unhashable type: 'numpy.ndarray'
```

## Fix
Install the strict reduction-axis normalizer on all three relevant modules:
- `pyrecest._backend`
- `pyrecest._backend._common`
- `pyrecest._backend.pytorch`

The normalizer converts every integer-like tuple element to a Python integer, while preserving the existing boolean-axis rejection and duplicate/out-of-range checks.

## Validation
- reproduced the old failure with `(np.asarray(0),)`
- targeted standalone validation confirms NumPy and Torch integer scalar tuple elements normalize correctly and boolean scalar elements remain rejected
- added backend-portable subprocess regression for the shared `min` path
- branch source and test files were re-fetched from GitHub after the commits to verify the final contents

Full repository pytest was not run locally because the available environment does not contain a PyRecEst checkout; GitHub CI should exercise the in-tree regression.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_reduction_axis_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_reduction_axis_contract.py
@@ -147,6 +147,30 @@ def _wrap_boolean_axis_dim_reduction(helper, torch_module):
     return reduction
 
 
+def _patch_reduction_normalizer(module, torch_module) -> None:
+    """Install the strict normalizer on one backend module."""
+
+    original_normalizer = getattr(module, "_normalize_reduction_axes", None)
+    if original_normalizer is None or getattr(
+        original_normalizer,
+        "_pyrecest_reduction_axis_bool_contract",
+        False,
+    ):
+        return
+
+    def _normalize_reduction_axes(axis, ndim_value):
+        return normalize_reduction_axes(axis, ndim_value, torch_module)
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4248, rebuilt on current `main` without overwriting the intervening flip/backend changes. #4248 merged successfully with shared-common normalization for NumPy and Torch integer scalar tuple axes and boolean-axis regression coverage.

PR #4222, opened 2026-07-12
Reject surplus coordinates in stacked product PDFs

FlorianPfaff:
## Summary
- validate `CartProdStackedDistribution.pdf()` inputs before splitting them across component distributions
- require single points and batches to match the exact stacked `input_dim`
- add regression coverage for surplus coordinates in both single-point and batched inputs

## Bug
`CartProdStackedDistribution.pdf()` sliced only the coordinates needed by its component distributions and never checked whether the full input had the expected dimension.

For a stacked model with `input_dim == 5`, both a six-dimensional point and an `(n, 6)` batch were accepted. The sixth coordinate was silently discarded, so inputs with an incorrect state layout produced plausible densities instead of failing fast.

## Fix
Validate that `xs` is either:
- a one-dimensional point of length `input_dim`, or
- a two-dimensional batch whose trailing dimension is `input_dim`

Only after that validation are component slices evaluated. Existing valid single-point and batched behavior is unchanged.

## Validation
- added regression coverage for surplus coordinates in single and batched inputs
- `python -m py_compile` passed for the modified implementation and new regression file
- branch is 2 commits ahead of current `main`, 0 behind
- final diff is limited to 5 implementation lines and one focused test module

The full repository/backend test matrix is delegated to GitHub Actions because this runtime cannot obtain a network checkout and does not provide the GitHub CLI.

### DIFF
--- a/src/pyrecest/distributions/cart_prod/cart_prod_stacked_distribution.py
+++ b/src/pyrecest/distributions/cart_prod/cart_prod_stacked_distribution.py
@@ -58,6 +58,11 @@ class CartProdStackedDistribution(AbstractCartProdDistribution):
 
     def pdf(self, xs):
         xs = asarray(xs)
+        if xs.ndim not in (1, 2) or xs.shape[-1] != self.input_dim:
+            raise ValueError(
+                "xs must be a one-dimensional point or a two-dimensional batch "
+                f"with trailing dimension {self.input_dim}, got shape {xs.shape}."
+            )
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4221. The canonical PR includes the surplus-coordinate validation from this PR and also fixes arbitrary leading batch dimensions by slicing along the trailing event axis. Closing the narrower duplicate.

PR #4237, opened 2026-07-12
Preserve nested batch shapes in partially wrapped PDFs

FlorianPfaff:
## Summary

- retain the original leading batch dimensions when normalizing `PartiallyWrappedNormalDistribution.pdf()` inputs
- reshape the final wrapped-normal density values back to that batch shape
- add a regression for input points with shape `(2, 3, 2)`

## Bug

`_as_2d_query_points()` accepts inputs with arbitrary leading batch dimensions as long as the trailing dimension matches the distribution dimension. It then flattens those points to `(n_eval, dim)` for the wrapped-normal computation, but `pdf()` returned the resulting one-dimensional vector directly.

For a valid input with shape `(2, 3, 2)`, the method therefore returned shape `(6,)` rather than one density per point with shape `(2, 3)`. This loses the caller's batch structure and can trigger broadcasting errors or misalign densities with structured inputs.

## Fix

Carry the intended density output shape through `_as_2d_query_points()` and reshape only the final summed density vector. Existing return contracts remain unchanged for:

- scalar inputs to one-dimensional distributions: `(1,)`
- one-dimensional sequences of scalar points: `(n,)`
- one multidimensional point: `(1,)`
- ordinary point matrices: `(n,)`

## Validation

- reproduced the old behavior: `(2, 3, 2)` input produced a flat `(6,)` result
- standalone regression confirms the patched result has shape `(2, 3)`
- every nested-batch entry matches evaluation of the corresponding point in isolation
- standalone compatibility checks confirm the established scalar, vector, and matrix output shapes are preserved
- branch is based directly on current `main`, two commits ahead and zero behind
- final diff is limited to the implementation and its existing test module

### DIFF
--- a/src/pyrecest/distributions/cart_prod/partially_wrapped_normal_distribution.py
+++ b/src/pyrecest/distributions/cart_prod/partially_wrapped_normal_distribution.py
@@ -86,7 +86,7 @@ def _validate_nonnegative_wrap_count(m) -> int:
 
 
 def _as_2d_query_points(xs, dim: int):
-    """Normalize scalar/vector/matrix query inputs to shape ``(n_eval, dim)``."""
+    """Normalize query inputs and retain the density output shape."""
     xs = array(xs)
     if xs.ndim == 0:
         if dim != 1:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4240, which implements the same batch-shape fix without the unrelated docstring edit and adds a focused flat-vs-nested regression. Closing this duplicate to keep one canonical change.

PR #4243, opened 2026-07-12
Stabilize wrapped-Cauchy PDFs for tiny gamma

FlorianPfaff:
## Summary

- evaluate the wrapped-Cauchy density with an algebraically equivalent half-angle `tanh(gamma / 2)` formulation
- avoid cancellation in `1 - exp(-gamma)` for tiny positive scale parameters
- add a regression at `gamma = 1e-20`

## Bug

`WrappedCauchyDistribution.pdf()` represented the concentration through `rho = exp(-gamma)` and then computed `1 - rho`.

For sufficiently small positive `gamma`, floating-point arithmetic rounds `exp(-gamma)` to exactly one. The numerator then becomes zero and the denominator is also zero at the mode, so a valid distribution returns `NaN`; away from the mode it returns zero density. For example, `gamma = 1e-20` produced `NaN` at `x = mu` and zero at ordinary off-mode evaluation points.

## Fix

Use the equivalent identity

```text
t = tanh(gamma / 2)
f(x) = t / (2 pi [sin^2((x-mu)/2) + t^2 cos^2((x-mu)/2)])
```

This form avoids subtracting nearly equal values. It remains stable in both limiting regimes: the sharply concentrated small-`gamma` case and the large-`gamma` uniform limit.

## Validation

- regression verifies finite, strictly positive densities for `gamma = 1e-20`
- regression checks the analytical modal density `1 / (2 pi tanh(gamma / 2))`
- syntax compilation passed for the modified implementation and new test
- standalone numerical checks passed for the tiny-`gamma` case and retained the `gamma = 1000` uniform limit
- branch is two commits ahead of current `main`, zero behind, and changes only the distribution implementation plus one focused regression file

### DIFF
--- a/src/pyrecest/distributions/circle/wrapped_cauchy_distribution.py
+++ b/src/pyrecest/distributions/circle/wrapped_cauchy_distribution.py
@@ -56,13 +56,18 @@ class WrappedCauchyDistribution(AbstractCircularDistribution):
     def pdf(self, xs):
         xs = _as_1d_input(xs)
         xs_centered = mod(xs - self.mu, 2 * pi)
-        rho = exp(-self.gamma)
-        one_minus_rho = 1.0 - rho
-        numerator = one_minus_rho * (1.0 + rho)
-        denominator = (
-            one_minus_rho**2 + 4.0 * rho * sin(xs_centered / 2.0) ** 2
-        )
-        return numerator / (2.0 * pi * denominator)
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4247, rebuilt on current `main` to preserve the merged wrapped-Cauchy validation and moment-order fixes. #4247 merged successfully with the stable half-angle PDF and tiny-`gamma` regression.

PR #4261, opened 2026-07-12
Prevent PyTorch input mutation during Bingham conjugation

FlorianPfaff:
## Summary

- use the PyRecEst backend copy operation in `BinghamFilter._conjugate()`
- prevent PyTorch tensors from being modified as a side effect of conjugation
- add a backend regression for input independence and the expected conjugate

## Bug

`BinghamFilter._conjugate()` used Python's `copy.copy()` before negating the vector part in place. NumPy arrays receive independent storage from that operation, but PyTorch tensors retain shared storage. On the PyTorch backend, calling `_conjugate(q)` therefore also negated `q[1:]` in the caller-owned input.

This backend-dependent side effect is especially risky when the argument is a view, as in the filter update path where matrix columns are passed into `_conjugate()`.

## Fix

Use `pyrecest.backend.copy()` for array duplication. The backend abstraction provides an independent array for mutable backends while preserving backend dtype and device semantics.

## Validation

- isolated NumPy/PyTorch reproduction confirms that Python shallow copies alias PyTorch tensor storage but not NumPy array storage
- regression verifies that the input remains unchanged and the returned quaternion is correctly conjugated
- JAX is skipped consistently with the existing `BinghamFilter` support restriction

Full backend and lint validation is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Superseded by merged #4307, which uses the active backend's owning copy in `BinghamFilter._conjugate()` and includes the PyTorch non-mutation regression. Closing this duplicate branch.

PR #4262, opened 2026-07-12
Return independent toroidal VM matrix shifts

FlorianPfaff:
## Summary

- deep-copy `ToroidalVMMatrixDistribution` before applying a shift
- prevent shifted distributions from sharing mutable `kappa` and `A` arrays with the source distribution
- add a regression that mutates shifted parameters and verifies the original remains unchanged

## Bug

`ToroidalVMMatrixDistribution.shift()` used `copy.copy(self)`. The returned object received a new `mu`, but its `kappa` and `A` attributes still referenced the exact mutable arrays owned by the source distribution.

Consequently, code that adjusted parameters on the shifted result could silently modify the original distribution. This violates the method's non-mutating return contract and can change later density evaluations of the source object without calling a method on it.

## Fix

Use `copy.deepcopy(self)` before replacing the shifted location. The shifted result now owns independent parameter arrays while preserving the existing location wrapping and numerical values.

## Validation

- regression captures the source `kappa` and `A`, mutates both arrays on the shifted distribution, and verifies the source remains unchanged
- JAX skips the mutation regression because its arrays are immutable
- branch comparison against `main`: two commits ahead, zero behind; one implementation-line replacement and one focused test addition

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Superseded by merged #4311, which applies the same deep-copy fix to `ToroidalVMMatrixDistribution.shift()` and includes the corresponding independence regression. Closing this duplicate branch.

PR #4266, opened 2026-07-12
Copy linear box-particle constructor arrays

FlorianPfaff:
## Summary

- copy box supports with the active backend's owning-copy operation
- copy explicit particle weights before storing them
- add NumPy and PyTorch regression coverage for separate and packed support inputs

## Bug

`LinearBoxParticleDistribution` converted constructor inputs to backend arrays and then stored them with Python's `copy.copy()`.

For NumPy arrays, a shallow copy happens to allocate independent data. For PyTorch tensors, however, `copy.copy()` preserves the same underlying storage. Mutating caller-owned `lower`, `upper`, or `w` tensors after construction could therefore silently change an existing distribution. The packed `(n_boxes, 2, dim)` input path was affected as well because its lower/upper slices shared the original tensor storage.

This could invalidate support ordering or weight normalization after those invariants had already been checked.

## Fix

Use `pyrecest.backend.copy()` for constructor-owned support arrays and explicit weights. This preserves backend dtype/device semantics while ensuring mutable backends do not retain caller storage. Existing uses of `copy.deepcopy()` for distribution-returning methods remain unchanged.

## Validation

- reproduced that `copy.copy()` keeps PyTorch tensor storage shared while NumPy arrays remain independent
- added an isolated backend-portable regression for NumPy and PyTorch
- regression covers separate `lower`/`upper` arrays, packed box arrays, and explicit normalized weights
- modified source and regression modules pass Python syntax compilation
- branch comparison against `main`: two commits ahead, zero behind; two files changed

The full backend and lint matrix is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Superseded by merged #4303, which implements the same backend-owning copies for linear box-particle supports and explicit weights with regression coverage. Closing this duplicate branch.

PR #4278, opened 2026-07-13
Normalize wrapped Cauchy scalar parameters

FlorianPfaff:
## Summary

- normalize accepted one-element wrapped-Cauchy `mu` and `gamma` arrays to zero-dimensional backend scalars
- preserve support for Python, NumPy, and backend scalar inputs
- add backend-aware regressions for parameter shapes, trigonometric moments, and numerical equivalence

## Bug

`WrappedCauchyDistribution` documents `mu` and `gamma` as scalar parameters, but `_validate_finite_scalar()` accepted arrays with shape `(1,)` and returned them unchanged.

That made scalar behavior depend on how numerically identical parameters were supplied. In particular, one-element parameter arrays propagated a singleton axis into `trigonometric_moment()`, which returned shape `(1,)` instead of a scalar. This can break scalar consumers and silently introduce an unintended batch dimension.

## Root cause

The validator checked that an input was either zero-dimensional or length one, but did not normalize the accepted length-one representation.

## Fix

Reshape accepted `(1,)` values to `()` during validation. Both `mu` and `gamma` now have stable scalar semantics while retaining the existing input acceptance policy.

## Validation

- added regression coverage that asserts scalar parameter and moment shapes through `pyrecest.backend.to_numpy()`
- added equivalence checks against ordinary scalar parameters for the PDF and second trigonometric moment
- inspected the final branch diff against current `main`; full backend execution is delegated to GitHub Actions

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Closing as duplicate of #4279. The latter is the preferred wrapped-Cauchy scalar normalization branch.

PR #4286, opened 2026-07-13
Stabilize discrete-state prior normalization

FlorianPfaff:
## Summary

- normalize finite discrete-state prior vectors after scaling by their maximum entry
- prevent overflow in the probability sum while preserving relative mass
- add a public forward/backward regression with extreme finite initial probabilities

## Bug

The compatibility-layer probability validator normalized priors with `values / values.sum()`. A valid vector can contain individually finite values whose sum overflows to infinity. For example, `[finfo.max, finfo.max / 2]` has a mathematically well-defined normalized ratio of `2:1`, but the direct floating-point sum is `inf`, so the implementation returned `[0, 0]`. The forward recursion then reported that the first emission had no prior-supported likelihood mass.

## Fix

Scale the non-negative vector by its largest entry before summing. All scaled entries are in `[0, 1]`, the relative probabilities are unchanged, and the positive-mass check remains explicit.

## Validation

- reproduced the pre-fix overflow and zero-vector normalization with NumPy
- verified scale-first normalization yields `[2/3, 1/3]`
- added `tests/filters/test_discrete_state_probability_overflow.py`
- branch comparison against `main`: two commits ahead, zero behind; 7 production-line changes and one focused regression file

The full repository test matrix is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Closing as superseded by #4405, the newer current-main implementation of stable discrete-state prior normalization.

PR #4287, opened 2026-07-13
Stabilize Merwe sigma-point scaling for small alpha

FlorianPfaff:
## Summary
- compute the Merwe sigma-point scale directly as `alpha**2 * (n + kappa)`
- avoid reconstructing it through `n + (alpha**2 * (n + kappa) - n)`
- reject non-finite or unrepresentable scales and weights with clear `ValueError`s
- add a regression for a valid small positive `alpha`

## Bug
`MerweScaledSigmaPoints` first formed

```python
lambda_ = alpha**2 * (n + kappa) - n
scale = n + lambda_
```

For sufficiently small but representable positive scales, subtracting and re-adding `n` causes catastrophic cancellation. With `n=2`, `kappa=0`, and `alpha=1e-9`, the mathematically correct scale is `2e-18`, but the current implementation rounds `lambda_` to `-2.0` and reconstructs `scale` as exactly `0.0`. Construction then fails with division by zero, and sigma-point generation would collapse the spread.

## Fix
Compute the positive scale directly and reuse that value for both weights and Cholesky scaling. This preserves the intended formula without the cancellation-prone round trip through `lambda`. The implementation also validates that the resulting scale and weights remain finite.

## Validation
- reproduced the pre-fix values: legacy scale `0.0`, direct scale `2e-18`
- verified the direct scale yields finite weights and sigma offsets with radius `sqrt(2e-18)`
- added backend-aware regression coverage for finite weights and nonzero sigma-point spread
- `python -m py_compile` passes for the modified source and regression test

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Closing as duplicate of #4289. The latter is the preferred Merwe sigma-point cancellation fix.

PR #4292, opened 2026-07-13
Reject non-finite sliding-window smoother weights

FlorianPfaff:
## Summary

- validate `SlidingWindowManifoldMeanSmoother.window_weights` for finiteness before sign and mass checks
- reject `NaN`, positive infinity, and negative infinity with a clear `ValueError`
- add focused regression coverage for all three non-finite cases

## Bug

`SlidingWindowManifoldMeanSmoother` checked only whether `window_weights` were negative and whether their sum was positive.

Those checks do not reject non-finite values:

- comparisons with `NaN` are false, so a vector containing `NaN` passed both checks
- a vector containing `+inf` had positive total mass and was accepted

When an active window was normalized, these values produced `NaN` weights (`NaN / NaN` or `inf / inf`) and consequently non-finite smoothed states.

## Fix

Require every configured window weight to be finite using the active backend before applying the existing non-negativity and positive-mass validation.

## Validation

- standalone numerical reproduction confirms the previous checks accept both `NaN` and `+inf` and normalization produces non-finite weights
- regression covers `NaN`, `+inf`, and `-inf`
- branch is two commits ahead of current `main` and zero behind
- final diff is limited to four source additions/one import-line adjustment and one focused test module

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Closing as an obsolete conflicted stack. This branch carries a large inherited commit chain on an old base, so the focused change cannot be reviewed or repaired safely in place. Recreate the small fix from current `main` if it is still missing.

PR #4294, opened 2026-07-13
Stabilize piecewise-constant weight normalization

FlorianPfaff:
## Summary

- scale finite nonnegative piecewise-constant weights by their maximum before normalization
- preserve relative interval mass while avoiding overflow in the raw mean and the `mean * 2π` denominator
- add a backend-aware regression using the active floating-point dtype's maximum finite value

## Bug

`PiecewiseConstantDistribution` normalized constructor weights with

```python
mean_weight = mean(w)
self.w = w / (mean_weight * 2.0 * pi)
```

Valid individually finite weights near the dtype maximum can overflow while computing the mean or while multiplying the mean by `2π`. The resulting infinite denominator collapses every stored density weight to zero. For example, `[finfo.max, finfo.max / 2]` should retain a 2:1 mass ratio, but the current implementation produces `[0, 0]`.

## Fix

Divide all weights by their maximum finite value first. The scaled values lie in `[0, 1]`, so their mean and subsequent normalization remain finite while the relative weights are unchanged.

## Validation

- isolated numerical reproduction: the previous formula yields `[0, 0]` for `[finfo.max, finfo.max / 2]`
- the scale-first formula yields densities `[2/(3π), 1/(3π)]` and integrates to one
- regression derives the maximum finite value from the active backend dtype and checks the expected normalized density
- modified source and regression modules pass Python bytecode compilation
- branch is two commits ahead of current `main`, zero behind; final diff is 8 production-line changes plus one focused 23-line regression

### DIFF
--- a/src/pyrecest/_backend/pytorch/signal.py
+++ b/src/pyrecest/_backend/pytorch/signal.py
@@ -5,7 +5,7 @@ _AXIS_TYPE_ERROR = "axes must be None, an integer, or a sequence of integers"
 
 
 def _is_boolean_scalar(axis):
-    return isinstance(axis, (bool, _np.bool_)) or (
+    return isinstance(axis, _np.bool_) or (
         isinstance(axis, _np.ndarray) and axis.shape == () and axis.dtype == _np.bool_
     )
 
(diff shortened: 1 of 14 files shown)


FlorianPfaff:
Closing as duplicate of #4295. The latter carries the same piecewise-constant normalization repair on the preferred branch.

PR #4370, opened 2026-07-14
Stabilize circular grid-to-Dirac weight conversion

FlorianPfaff:
## Summary
- scale circular grid weights by their largest finite positive entry before constructing a `CircularDiracDistribution`
- preserve relative weights while avoiding overflow in direct normalization
- add a NumPy regression for two maximum finite `float64` grid weights

## Root cause
`CircularDiracDistribution.from_distribution()` normalized grid weights with `weights / sum(weights)`. For individually finite large weights, the sum can overflow to infinity. Dividing by that infinite total collapses every weight to zero, and the Dirac constructor then rejects the otherwise valid conversion because the resulting weights have no positive mass.

## Impact
Deterministic conversion from a circular grid distribution now succeeds for extreme but finite weights and preserves their relative probabilities. Invalid inputs such as non-finite, negative, empty, or zero-mass weights are still rejected by the existing Dirac-weight validation.

## Validation
- reproduced the pre-fix behavior with two `np.finfo(float).max` weights: direct normalization produces `[0, 0]`
- verified scale-before-normalize produces `[0.5, 0.5]` with unit total mass
- `python -m py_compile` passes for the patched conversion code
- branch is two commits ahead of `main` and zero behind; only the conversion and focused regression test changed

The repository's full backend and lint matrices are delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/distributions/circle/circular_dirac_distribution.py
+++ b/src/pyrecest/distributions/circle/circular_dirac_distribution.py
@@ -1,4 +1,11 @@
-from pyrecest.backend import isfinite, max as backend_max, ones, reshape
+from pyrecest.backend import all as backend_all
+from pyrecest.backend import (
+    isfinite,
+    max as backend_max,
+    ones,
+    reshape,
+    sum as backend_sum,
+)
 
 from ..hypertorus.hypertoroidal_dirac_distribution import HypertoroidalDiracDistribution
 from .abstract_circular_distribution import AbstractCircularDistribution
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
CI root cause and fix:

The regression test imported `CircularGridDistribution` from `pyrecest.distributions`, but that class is not exported by the package-level distribution namespace. Pytest therefore failed during test-module collection on every backend, before the NumPy-only skip decorator could take effect.

Fixed by importing the class from `pyrecest.distributions.circle.circular_grid_distribution`.

I also completed the scale-before-normalize implementation so valid finite nonnegative grid weights are fully normalized before constructing the Dirac distribution. This preserves invalid-input validation while avoiding the new `Weights are not normalized` runtime warning.

Current head: `18f749fd4a9ef1b09131ce0b74f0d2e1a8c2147a`

Local focused validation:
- NumPy: 5 passed, no warnings
- PyTorch: 4 passed, 1 NumPy-specific test skipped
- JAX: all 5 tests collect successfully

The complete GitHub Actions matrix has been retriggered.

PR #4383, opened 2026-07-14
Stabilize sliding-window smoother weights

FlorianPfaff:
## Summary

- validate smoothing window weights are finite before use
- scale finite nonnegative weights by their largest entry during construction
- preserve relative weights when a direct floating-point sum would overflow
- add a backend-aware regression through the public sliding-window smoother API

## Root cause

`SlidingWindowManifoldMeanSmoother` stored raw `window_weights` and summed them directly both during construction and for each active edge window. Individually finite weights can still have an infinite sum. For example, `[finfo.max, finfo.max / 2]` has the well-defined ratio `2:1`, but direct normalization produces `[0, 0]`. The subsequent Dirac distribution then rejects the zero-mass weights instead of returning the weighted manifold mean.

## Fix

After shape, finiteness, nonnegativity, and positive-mass validation, scale the complete weight vector by its largest entry. The stored values are bounded by one and preserve the original ratios, so all later truncated-window sums remain finite and can be normalized safely.

## Validation

- reproduced the pre-fix overflow: maximum finite weights sum to `inf` and direct normalization collapses to zero
- verified scale-first normalization preserves the expected `2/3, 1/3` ratio
- added a trailing-window regression where states `[0, 3]` smooth to `1` under extreme `2:1` weights
- compared the branch with `main`: two commits ahead, zero behind; only the smoother implementation and its existing test module changed

The full backend and lint matrix is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/smoothers/sliding_window_manifold_mean_smoother.py
+++ b/src/pyrecest/smoothers/sliding_window_manifold_mean_smoother.py
@@ -6,8 +6,17 @@ from collections.abc import Callable, Sequence
 from functools import partial
 from operator import index as operator_index
 
+from pyrecest.backend import all as backend_all
 from pyrecest.backend import any as backend_any
-from pyrecest.backend import asarray, concatenate, ndim, stack, sum
+from pyrecest.backend import (
+    asarray,
+    concatenate,
+    isfinite,
+    max as backend_max,
+    ndim,
+    stack,
+    sum,
+)
 from pyrecest.distributions import (
     AbstractHypercylindricalDistribution,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Current `main` already contains a stronger implementation of this fix: window weights are validated as finite/non-negative and every active edge-window slice is scale-normalized through two square-root-sized divisions before summation. That covers the extreme finite-weight overflow addressed here while preserving truncated-window behavior. Closing this now-conflicted branch as superseded rather than reintroducing the older construction-time scaling approach.

PR #4420, opened 2026-07-15
Fix particle diagnostics normalization overflow

### LINKED ISSUE #4415: Stabilize ParticleDiagnostics normalization for extreme finite weights
## Bug

`ParticleDiagnostics.from_weights()` normalizes nonnegative finite weights by summing them at their original scale. For valid inputs near the `float64` limit, the sum can overflow to infinity even though all entries and their relative probabilities are well defined.

For example:

```python
maximum = np.finfo(float).max
diagnostics = ParticleDiagnostics.from_weights([maximum, maximum / 2.0])
```

The current normalization computes an infinite total and therefore produces `[0.0, 0.0]`. The reported effective sample size and entropy are both zero instead of the correct values:

- normalized weights: `[2/3, 1/3]`
- ESS: `1.8`
- entropy: approximately `0.6365141683`

## Suggested fix

After validating finiteness, clip negative diagnostic weights as today, divide by the largest positive entry, and normalize the bounded scaled vector. Add a focused regression through `ParticleDiagnostics.from_weights()`.

FlorianPfaff:
## Summary
- normalize clipped particle weights after scaling by their maximum finite value
- preserve relative probabilities when the original finite sum would overflow
- add a regression for weights near the `float64` limit, checking ESS and entropy

## Testing
- `python -m py_compile` on the modified source and test files
- isolated `pytest -q tests/test_diagnostics.py` (`6 passed`)

Fixes #4415

### DIFF
--- a/src/pyrecest/diagnostics.py
+++ b/src/pyrecest/diagnostics.py
@@ -176,10 +176,12 @@ def _normalized_nonnegative_weights(values: list[float]) -> list[float]:
     if any(not isfinite(value) for value in values):
         raise ValueError("Particle weights must be finite.")
     nonnegative_values = [max(0.0, value) for value in values]
-    total = sum(nonnegative_values)
-    if total <= 0.0:
+    scale = max(nonnegative_values, default=0.0)
+    if scale <= 0.0:
         return [0.0 for _ in nonnegative_values]
-    return [value / total for value in nonnegative_values]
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant with #4419, which implements the same particle-diagnostics overflow normalization fix, includes focused regression coverage, and is already ready with auto-merge enabled.

PR #4429, opened 2026-07-15
Stabilize particle diagnostics weight normalization

### LINKED ISSUE #4415: Stabilize ParticleDiagnostics normalization for extreme finite weights

FlorianPfaff:
## Summary

- normalize finite nonnegative particle diagnostic weights after scaling by their largest entry
- preserve relative mass when the caller-provided weight sum overflows
- add a focused regression for finite weights near the `float64` limit

## Bug

`ParticleDiagnostics.from_weights()` summed weights at their original scale. Inputs such as `[max_float, max_float / 2]` contain only finite values and define valid relative probabilities, but their raw sum is infinite. Dividing by that total produced `[0.0, 0.0]`, so effective sample size and entropy were both reported as zero.

## Fix

Clip negative values as before, scale the nonnegative vector by its largest positive entry, and then normalize the bounded scaled vector. At least one scaled value is exactly one whenever positive mass exists, so the normalization sum cannot overflow or underflow to zero.

## Validation

- implementation diff is limited to the normalization helper
- focused regression verifies that every input weight is finite while the unscaled sum is infinite
- reproduced normalized weights `[2/3, 1/3]`, ESS `1.8`, and entropy `0.6365141683`

Fixes #4415

### DIFF
--- a/.github/ci-refresh/4429.txt
+++ b/.github/ci-refresh/4429.txt
@@ -1,1 +1,0 @@
-Temporary CI refresh marker.
--- a/.github/ci-refresh/4429.txt
+++ b/.github/ci-refresh/4429.txt
@@ -1,0 +1,1 @@
+Temporary CI refresh marker.
--- a/src/pyrecest/diagnostics.py
+++ b/src/pyrecest/diagnostics.py
@@ -176,10 +176,12 @@ def _normalized_nonnegative_weights(values: list[float]) -> list[float]:
     if any(not isfinite(value) for value in values):
         raise ValueError("Particle weights must be finite.")
     nonnegative_values = [max(0.0, value) for value in values]
-    total = sum(nonnegative_values)
-    if total <= 0.0:
+    scale = max(nonnegative_values, default=0.0)
+    if scale <= 0.0:
         return [0.0 for _ in nonnegative_values]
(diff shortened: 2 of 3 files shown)


FlorianPfaff:
Closing as low-priority extreme-value hardening. The motivating particle-weight case is at the floating-point limit, while the branch remains red in the optional backend matrix. This does not justify additional cross-backend normalization machinery without a realistic failure report.

PR #4432, opened 2026-07-15
Normalize circular grid weights without overflow

FlorianPfaff:
## Summary

- normalize circular S3F grid weights after scaling by their largest entry
- preserve relative probability mass when a finite raw weight sum overflows
- add a focused regression using finite values near the `float64` limit

## Bug

`grid_probability_masses()` summed nonnegative weights at their caller-provided scale. Inputs such as `[max_float, max_float / 2, max_float / 4]` are individually finite and define valid relative probabilities, but their raw sum overflows to infinity. The previous implementation then divided every finite weight by infinity and silently returned all-zero masses. This also corrupted downstream circular weighted means.

## Fix

Reject empty or all-zero inputs, scale all weights by their largest positive entry, and normalize the scaled values. Scaling preserves all relative masses and zero support while ensuring the summation remains finite.

## Validation

- isolated reproduction confirms the old calculation returns all zeros for an overflowing finite sum
- the regression verifies the fixed result equals the safely scaled probability vector and sums to one
- branch comparison is limited to the normalization helper and one focused regression file

The full configured test and lint suite is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/filters/relaxed_s3f_circular.py
+++ b/src/pyrecest/filters/relaxed_s3f_circular.py
@@ -35,6 +35,7 @@ from pyrecest.backend import (
     sin,
     stack,
 )
+from pyrecest.backend import max as backend_max
 from pyrecest.backend import sum as backend_sum
 from pyrecest.backend import (
     zeros,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-priority numerical hardening. This overlaps several concurrent maximum-finite-value normalization PRs, while the full backend test matrix is red. Rather than carry multiple narrowly scoped variants, this should be revisited as one consolidated, backend-tested numerical-stability change.

PR #4433, opened 2026-07-15
Avoid overflow in covariance symmetrization

FlorianPfaff:
## Summary

- add an overflow-safe matrix symmetrization helper
- use it in `symmetrize_matrix`, `nearest_symmetric_psd`, and `jittered_cholesky`
- add regression coverage for maximum-finite diagonal covariance matrices

## Bug

The shared numerical helpers evaluated `0.5 * (A + A.T)`. For finite entries near `np.finfo(float).max`, the intermediate addition overflows before the factor of one half is applied. A valid finite covariance can therefore become infinite inside symmetrization, PSD projection, or Cholesky preparation.

For example, a diagonal covariance with maximum-finite entries is mathematically unchanged by symmetrization, but the old expression produces `inf` on its diagonal.

## Fix

Keep the ordinary averaging expression, then recompute only entries where finite operands produced a non-finite result as `0.5 * A + 0.5 * A.T`. This avoids the overflowing intermediate sum while preserving the existing behavior for ordinary values.

## Validation

- reproduced the old overflow with a maximum-finite diagonal matrix
- verified the patched helper keeps the matrix and Cholesky factor finite
- added a regression covering symmetrization, nearest-PSD projection, and jittered Cholesky

Full repository tests were not runnable in this environment because local GitHub checkout access is unavailable; GitHub Actions should provide authoritative validation.

### DIFF
--- a/src/pyrecest/numerics.py
+++ b/src/pyrecest/numerics.py
@@ -146,11 +146,25 @@ def _raise_if_not_square_matrix(matrix: np.ndarray, name: str = "matrix") -> Non
         raise ShapeError(f"Expected a square matrix, got shape {matrix.shape}.")
 
 
+def _symmetrized(matrix: np.ndarray) -> np.ndarray:
+    """Average a matrix with its transpose without overflowing finite entries."""
+    with np.errstate(over="ignore"):
+        symmetrized = 0.5 * (matrix + matrix.T)
+    overflowed = (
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as a high-risk, low-priority extreme-value hardening change. It modifies shared covariance numerics for inputs near the floating-point maximum, overlaps the current overflow-hardening cluster, and the full backend matrix is red. This is better revisited in a consolidated numerical-stability PR with explicit cross-backend tests.

PR #4435, opened 2026-07-15
Avoid overflow in hypertoroidal grid-to-Dirac conversion

FlorianPfaff:
## Summary

- remove redundant raw-sum normalization from `HypertoroidalDiracDistribution.from_distribution()`
- delegate normalization to the Dirac constructor's existing overflow-safe max-scaled implementation
- add a focused regression for finite grid weights whose direct sum overflows

## Bug

Grid-backed hypertoroidal distributions were normalized with `weights / sum(weights)` before constructing the Dirac approximation. A valid finite vector such as `[max_float, max_float / 2]` has a raw sum of infinity, so the pre-normalization produced `[0, 0]`. The Dirac constructor then rejected the result as having no positive mass, even though the original weights define the valid probability vector `[2/3, 1/3]`.

## Fix

Pass the raw grid values to `HypertoroidalDiracDistribution`. `AbstractDiracDistribution` already validates and normalizes weights after scaling by their maximum, preserving relative mass without an overflowing intermediate sum.

## Impact

Deterministic conversion from grid-backed hypertoroidal distributions now works for all finite nonnegative weight scales accepted by the Dirac distribution itself. Ordinary-scale conversions are unchanged.

## Validation

- reproduced the old overflow: direct normalization returns `[0, 0]`
- verified max-scaled normalization returns `[2/3, 1/3]`
- added a regression that also verifies conversion does not fall back to sampling
- syntax-compiled the regression module
- branch is two commits ahead of `main`, zero behind, with changes limited to one implementation line and one focused test file

The full backend and lint matrices are delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/distributions/hypertorus/hypertoroidal_dirac_distribution.py
+++ b/src/pyrecest/distributions/hypertorus/hypertoroidal_dirac_distribution.py
@@ -89,7 +89,7 @@ class HypertoroidalDiracDistribution(
         get_grid = getattr(distribution, "get_grid", None)
         if hasattr(distribution, "grid_values") and callable(get_grid):
             weights = reshape(distribution.grid_values, (-1,))
-            weights = weights / sum(weights)
+            # The Dirac constructor performs overflow-safe normalization.
             return HypertoroidalDiracDistribution(
                 get_grid(), weights, dim=distribution.dim
             )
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant low-priority overflow hardening. This relies on the Dirac constructor's existing normalization and overlaps the concurrent maximum-finite-value fixes, while the full backend matrix is red. A future consolidated numerical-stability PR can cover these normalization paths consistently.

PR #4436, opened 2026-07-15
Avoid overflow in box center and half-width calculations

FlorianPfaff:
## Summary

- compute box centers by scaling each endpoint before addition
- compute half-widths by scaling each endpoint before subtraction
- add backend-aware regressions for finite bounds near the active floating-point limit

## Bug

`LinearBoxParticleDistribution.centers()` evaluated `0.5 * (lower + upper)`. For same-sign finite bounds near the dtype maximum, the intermediate endpoint sum overflows even when the mathematical midpoint is finite. For example, `[max / 2, max]` produced an infinite center, which also contaminated `d`, `mean()`, and `mode()`.

`half_widths()` had the analogous issue for opposite extreme bounds: `0.5 * (upper - lower)` overflowed for `[-max, max]`, although the half-width itself is exactly `max` and representable.

## Fix

Scale each finite endpoint before combining it:

- center: `0.5 * lower + 0.5 * upper`
- half-width: `0.5 * upper - 0.5 * lower`

This avoids the overflowing intermediate sum or difference while preserving the existing result for ordinary inputs.

## Validation

- reproduced both failures with NumPy `float32` and `float64`
- verified the revised expressions remain finite and return the expected values
- added regressions that derive the limit from the active PyRecEst backend dtype

The repository's configured GitHub Actions checks provide the full-suite validation.

### DIFF
--- a/src/pyrecest/distributions/nonperiodic/linear_box_particle_distribution.py
+++ b/src/pyrecest/distributions/nonperiodic/linear_box_particle_distribution.py
@@ -130,7 +130,7 @@ class LinearBoxParticleDistribution(AbstractLinearDistribution):
 
     def centers(self):
         """Return all box centers with shape ``(n_boxes, dim)``."""
-        return 0.5 * (self.lower + self.upper)
+        return 0.5 * self.lower + 0.5 * self.upper
 
     def widths(self):
         """Return side lengths of all boxes."""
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-priority extreme-bound hardening. The change targets representable results at near-maximum floating-point bounds and overlaps the current numerical-overflow cluster, while the full backend matrix is red. This is better handled in one consolidated, cross-backend numerical-stability change.

PR #4438, opened 2026-07-15
Preserve nonuniform weights in Daum-Huang updates

FlorianPfaff:
## Bug

`DaumHuangParticleFlowFilter.update_model()` supplies the current particle weights to the Gaussian transport, but the existing implementation reconstructs the transported `LinearDiracDistribution` with uniform weights. A nonuniform prior therefore produces an output distribution whose weighted moments no longer match the moments used to construct the transport.

## Fix

- retain the existing implementation unchanged in a private backing module
- keep the public module/API stable through a compatibility wrapper
- restore the normalized prior weights on the transported particles for both EDH and LEDH filters
- add a focused regression checking weight preservation and the resulting posterior mean/covariance against analytical Gaussian bridge moments

## Validation

- exact source blob from `main` was used as the backing implementation
- focused NumPy regression passes locally (`1 passed`)
- syntax compilation passes for the wrapper, backing module, and regression
- repository CI is running on the final three-file tree

No temporary workflow remains in the PR diff.


FlorianPfaff:
Closing as hard to maintain in its current form. The behavioral issue is real, but this PR duplicates the complete 749-line Daum–Huang implementation into a private module and wraps it to change one weight assignment. That creates unnecessary code duplication and review risk. The fix should be reintroduced later as a minimal in-place change after the shared test baseline is green.

PR #4439, opened 2026-07-15
Normalize IMM probabilities without overflow

FlorianPfaff:
## Summary

- normalize IMM transition rows after scaling each row by its largest entry
- normalize mode probabilities after scaling by their largest entry
- add a focused regression for finite values whose raw sums overflow

## Bug

`InteractingMultipleModelFilter` normalized transition rows and mode probabilities by summing values at their original scale. Valid finite inputs near the `float64` maximum can have an infinite raw sum. Dividing by that sum silently produced all-zero transition rows or all-zero model probabilities.

For example, the finite relative weights `[max_float, max_float / 2]` define probabilities `[2/3, 1/3]`, but the previous normalization returned `[0, 0]`.

## Fix

Scale each nonnegative probability vector by its largest positive entry before summing and normalizing. This preserves relative mass while keeping the intermediate sums finite. Transition matrices apply the same normalization independently to every row.

## Validation

- reproduced the old all-zero result with NumPy `float64` maximum-finite values
- verified the patched normalization returns the expected `2/3` and `1/3` probabilities with unit sums
- added regression coverage for both transition-row and mode-probability normalization
- changes are limited to the IMM implementation and one focused test file

The repository's configured GitHub Actions checks provide the full-suite validation.

### DIFF
--- a/src/pyrecest/filters/interacting_multiple_model_filter.py
+++ b/src/pyrecest/filters/interacting_multiple_model_filter.py
@@ -21,6 +21,7 @@ from pyrecest.backend import (
     isfinite,
     linalg,
     log,
+    max as backend_max,
     ones,
     outer,
     pi,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant low-priority maximum-float normalization hardening. It overlaps the other overflow-scale probability PRs already closed in this triage and increases backend surface area in the IMM core. Revisit this as part of one consolidated numerical-stability change after the shared test baseline is green.

PR #4445, opened 2026-07-15
Reject nonfinite jittered Cholesky factors

FlorianPfaff:
## Bug

`jittered_cholesky()` accepted finite inputs and finite retry controls, but did not validate the matrix produced after adding diagonal jitter or the returned Cholesky factor.

For a matrix such as `diag(max_float / 2, -1)` with `initial_jitter=max_float`, the raw factorization fails as expected. Adding the first jitter then overflows one diagonal entry to `inf`. NumPy can still return a factor containing `inf`, and the helper returned that factor as a successful result.

## Fix

- form each jittered candidate under a local floating-point error state
- stop when adding jitter produces nonfinite values
- reject a nonfinite factor even if the linear algebra routine does not raise
- preserve the existing retry schedule and failure exception
- add a focused regression that reproduces the overflow path

## Validation

- reproduced the previous nonfinite factor locally with NumPy
- verified the patched logic raises `NumericalStabilityError` for the overflow case
- verified the ordinary singular-PSD jitter path still returns a finite factor
- branch is based on current `main` and changes only `src/pyrecest/numerics.py` plus one focused regression test

### DIFF
--- a/.github/ci-refresh/4445.txt
+++ b/.github/ci-refresh/4445.txt
@@ -1,1 +1,0 @@
-Temporary CI refresh marker.
--- a/.github/ci-refresh/4445.txt
+++ b/.github/ci-refresh/4445.txt
@@ -1,0 +1,1 @@
+Temporary CI refresh marker.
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing as low-priority extreme-value hardening. The motivating failure requires Cholesky inputs/jitter at near-maximum floating-point scale, and the branch remains red in the optional backend matrix. A backend-neutral implementation is not justified by a realistic use case.

PR #4448, opened 2026-07-15
Stabilize SO(3) chordal weight normalization

FlorianPfaff:
## Bug

`SO3ChordalMeanSmoother._normalize_weight_vector()` summed finite nonnegative weights at their original scale. For weights near the active floating-point maximum, the sum could overflow to infinity. Dividing by that total then collapsed every normalized weight to zero, so the projected chordal mean could become an unrelated rotation.

For example, weights proportional to `3:1` at the `float64` limit produced the identity rotation instead of the same weighted mean as ordinary `[3.0, 1.0]` weights.

The same risk applied when finite sample and kernel weights were multiplied during local smoothing.

## Fix

- scale every weight vector by its largest positive entry before summation
- normalize the bounded scaled weights when a probability vector is needed
- retain the bounded relative weights for kernel/sample weighting so their products cannot overflow
- preserve existing validation for length, finiteness, nonnegativity, and positive mass

## Validation

- reproduced overflow of the old raw weight sum and the resulting identity rotation
- verified scaled normalization yields `[0.75, 0.25]` for the extreme finite `3:1` case
- added a backend-aware regression comparing extreme weights with ordinary `[3.0, 1.0]` weights
- branch is 2 commits ahead of `main`, 0 behind, and changes only the smoother plus the focused regression

### DIFF
--- a/.github/ci-refresh/4448.txt
+++ b/.github/ci-refresh/4448.txt
@@ -1,1 +1,0 @@
-Temporary CI refresh marker.
--- a/.github/ci-refresh/4448.txt
+++ b/.github/ci-refresh/4448.txt
@@ -1,0 +1,1 @@
+Temporary CI refresh marker.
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing as low-priority extreme-value hardening. The motivating case requires SO(3) chordal weights near the floating-point limit, while the branch remains red in the optional backend matrix. This is not worth carrying additional normalization complexity without a realistic failing workload.

PR #4449, opened 2026-07-15
Stabilize large finite time-offset errors

FlorianPfaff:
## Bug

`time_offset_error_summary()` computed residual magnitudes with `np.linalg.norm` at the original scale. Large but finite residual components could overflow during squaring, producing `inf`. `_error_stats()` then filtered those values out as non-finite and reported zero valid samples.

The same overflow pattern affected mean, standard-deviation, and RMSE reductions, including aggregated time-offset sweeps. For example, two valid residuals equal to `0.75 * np.finfo(float).max` were reported as an empty result even though each residual and each expected summary metric is representable.

## Fix

- scale residual rows before computing Euclidean norms
- compute means, standard deviations, and RMS values in normalized coordinates before restoring the original scale
- scale aggregation weights and pooled-variance calculations to avoid intermediate overflow
- preserve existing behavior for zero, empty, ordinary-scale, and genuinely non-representable results

## Validation

- reproduced the previous `inf` norm and zero-count summary
- added end-to-end regressions for large finite residual summaries and aggregated sweeps
- focused regression tests: 2 passed
- syntax compilation passed for the implementation and regression test
- compatibility checks passed for offset-grid generation, nearest-time matching, existing aggregate semantics, and recovery of a known offset

### DIFF
--- a/.github/ci-refresh/4449.txt
+++ b/.github/ci-refresh/4449.txt
@@ -1,1 +1,0 @@
-Temporary CI refresh marker.
--- a/.github/ci-refresh/4449.txt
+++ b/.github/ci-refresh/4449.txt
@@ -1,0 +1,1 @@
+Temporary CI refresh marker.
--- a/src/pyrecest/calibration/time_offset.py
+++ b/src/pyrecest/calibration/time_offset.py
@@ -205,6 +205,76 @@ def _as_nonnegative_summary_count(value: Any, name: str) -> float:
     return result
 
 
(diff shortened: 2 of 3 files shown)


FlorianPfaff:
Closing as low-value numerical hardening. The change adds substantial aggregation machinery for residuals near the `float64` maximum, far outside realistic calibration-error scales, while the full backend matrix remains red. The maintenance cost outweighs the practical benefit.

PR #4452, opened 2026-07-15
Normalize JAX FFT shift scalar axes

FlorianPfaff:
## Bug

The JAX `fftshift` and `ifftshift` wrappers forwarded `axes` directly to `jax.numpy.fft`. Python integer axes worked, but equivalent NumPy/JAX integer scalars and zero-dimensional integer arrays did not. For example, `axes=np.int64(1)`, `axes=np.array(1)`, and `axes=jnp.array(1)` were treated as iterables by JAX and raised `TypeError` instead of selecting axis 1.

Boolean axes also reached JAX and produced backend-dependent low-level errors rather than the explicit integer-axis validation used by the neighboring FFT wrappers.

## Fix

- add a shift-specific axis normalizer that accepts Python, NumPy, and JAX integer scalars
- convert zero-dimensional integer arrays to Python `int`
- preserve and normalize valid axis sequences
- reject boolean scalar/array axes with a stable `TypeError`
- route both `fftshift` and `ifftshift` through the normalizer

## Regression coverage

Added parameterized tests for both shift directions covering:

- `np.int64(1)`
- `np.array(1)`
- `jnp.array(1)`
- Python, NumPy, and JAX boolean axes

## Validation

- `python -m py_compile` on the modified wrapper and regression test
- isolated JAX smoke checks comparing scalar-array axes with the equivalent Python integer for both `fftshift` and `ifftshift`

The full repository suite was not run locally because this environment cannot clone the repository; GitHub Actions should run the repository matrix.

### DIFF
--- a/src/pyrecest/_backend/jax/fft.py
+++ b/src/pyrecest/_backend/jax/fft.py
@@ -11,6 +11,11 @@ _BOOLEAN_FFT_AXIS_ERROR = "axis must be an integer, not boolean"
 _BOOLEAN_FFT_LENGTH_ERROR = "n must be an integer length, not boolean"
 _FFT_SHAPE_SEQUENCE_ERROR = "s must be None or a sequence of integer lengths"
 _FFT_AXES_SEQUENCE_ERROR = "axes must be None or a sequence of integer axes"
+_FFT_SHIFT_AXES_ERROR = (
+    "axes must be None, an integer axis, or a sequence of integer axes"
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as a low-value backend edge case in its current form. The JAX-specific regression remains red, so this should be resubmitted only with a passing focused JAX path and without relying on the broader failing matrix.

PR #4453, opened 2026-07-15
Reject overflowing reliability covariance products

FlorianPfaff:
## Bug

`scale_covariance_by_reliability()` validated the input covariance and computed a finite reliability scale, but returned `covariance * scale` without validating the product. Finite inputs could therefore produce an infinite covariance when the multiplication overflowed.

For example, scaling `[[np.finfo(float).max]]` by the valid scale `2.0` returned `[[inf]]`. The public low-level helper leaked that invalid covariance directly, while the high-level reliability path only failed later through generic result validation.

## Fix

- perform covariance scaling under a local NumPy error state
- reject any non-finite scaled covariance with a clear `ValueError`
- preserve the existing finite `max_scale` mechanism as a way to keep the product representable

## Tests

- regression for both `scale_covariance_by_reliability()` and `apply_measurement_reliability()`
- regression showing `max_scale=1.0` safely preserves a covariance at the floating-point maximum
- local `py_compile` for the modified module and new test file
- isolated targeted validation of normal scaling, overflow rejection, high-level propagation, and the safe bounded path

### DIFF
--- a/src/pyrecest/tracking/measurement_reliability.py
+++ b/src/pyrecest/tracking/measurement_reliability.py
@@ -187,7 +187,13 @@ def scale_covariance_by_reliability(
         exponent=exponent,
         max_scale=max_scale,
     )
-    return covariance * scale, scale
+    with np.errstate(over="ignore", invalid="ignore"):
+        scaled_covariance = covariance * scale
+    if not np.isfinite(scaled_covariance).all():
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-priority numerical hardening. The failure requires covariance values at the floating-point maximum, while the proposed NumPy-specific validation breaks the optional backend matrix. A backend-neutral redesign is disproportionate to the practical value of this edge case.

PR #4454, opened 2026-07-15
Stabilize IMM Gaussian likelihood evaluation

FlorianPfaff:
## Bug

`InteractingMultipleModelFilter._log_linear_measurement_likelihood()` computed `det(S)` before taking its logarithm. For a valid positive-definite innovation covariance at an extreme scale, the determinant can underflow to zero or overflow to infinity even though the log-determinant remains finite.

That could incorrectly reject a valid covariance or collapse a model likelihood to `-inf`.

## Fix

- evaluate the covariance normalization with `linalg.slogdet`
- preserve the positive-definiteness check through the determinant sign
- reject nonfinite log-determinants explicitly
- add regressions for covariance scales `1e-200` and `1e200`

## Validation

- reconstructed and verified the source against the current Git blob before editing
- parsed the patched module and regression test successfully
- checked the extreme-scale likelihood calculation directly: both cases remain finite and match the signed-log-determinant formula
- verified the branch diff contains only the IMM source change and its regression test

Normal repository CI is triggered by this pull request; it may remain queued while the account-wide GitHub Actions backlog clears.


FlorianPfaff:
This draft was superseded after the fix branch was rebuilt cleanly from current `main`. The final two-file change is in PR #4456.

PR #4456, opened 2026-07-15
Stabilize IMM Gaussian likelihood evaluation

FlorianPfaff:
## Bug

`InteractingMultipleModelFilter._log_linear_measurement_likelihood()` computed `det(S)` before taking its logarithm. For a valid positive-definite innovation covariance at an extreme scale, the determinant can underflow to zero or overflow to infinity even though the log-determinant remains finite.

That could incorrectly reject a valid covariance or collapse a model likelihood to `-inf`.

## Fix

- evaluate the covariance normalization with `linalg.slogdet`
- preserve the existing covariance-validity guard through the determinant sign
- reject nonfinite log-determinants explicitly
- add regressions for covariance scales `1e-200` and `1e200`

## Validation

- reconstructed and verified the source against the current Git blob before editing
- parsed the patched module and regression test successfully
- executed the committed module and new regression in an isolated package harness: `2 passed`
- verified the branch is two commits ahead of `main`, not behind, with only the IMM source and regression-test files changed

Normal repository CI is triggered by this pull request; checks are currently queued in the account-wide GitHub Actions backlog.

### DIFF
--- a/src/pyrecest/filters/interacting_multiple_model_filter.py
+++ b/src/pyrecest/filters/interacting_multiple_model_filter.py
@@ -612,12 +612,13 @@ class InteractingMultipleModelFilter(AbstractFilter, EuclideanFilterMixin):
             @ measurement_matrix.T
             + meas_noise
         )
-        det_value = float(linalg.det(innovation_covariance))
-        if det_value <= 0.0:
+        determinant_sign, logdet = linalg.slogdet(innovation_covariance)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing this draft as hard to justify in its current form. The change alters core IMM likelihood evaluation, while the PR-specific test matrix fails across the relevant configurations. A future version should first establish backend-compatible `slogdet` behavior and a narrower regression.

PR #4457, opened 2026-07-16
Stabilize grid integral averaging

FlorianPfaff:
## Bug

`AbstractGridDistribution.integrate()` computed the grid average with the backend's raw `mean`. For finite values at extreme scale, the underlying reduction can overflow before cancellation occurs.

For example, grid values proportional to `[max_float, max_float, -max_float, -max_float]` have an exact mean and integral of zero, but NumPy's raw reduction overflows to `inf`. `normalize_in_place()` then rejects the distribution as having a nonfinite integral even though the represented integral is finite.

## Fix

- scale nonempty finite grid values by their maximum absolute magnitude before averaging
- restore the original scale after the bounded mean reduction
- retain the previous behavior for empty, zero-valued, and nonfinite grids
- add a backend-aware regression for the cancelling extreme-value case

## Impact

Grid integration and normalization no longer fail solely because an intermediate sum overflows when large positive and negative finite values cancel.

## Validation

- reproduced the old NumPy result: raw mean becomes `inf`
- verified the scaled reduction returns the exact finite result `0.0`
- branch is 3 commits ahead of `main`, 0 behind
- diff is limited to the abstract grid integration helper and its focused test module

Full repository CI is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/distributions/abstract_grid_distribution.py
+++ b/src/pyrecest/distributions/abstract_grid_distribution.py
@@ -83,6 +83,8 @@ class AbstractGridDistribution(AbstractDistributionType):
             raise NotImplementedError(
                 "Custom integration boundaries are currently not supported."
             )
+        if prod(self.grid_values.shape) == 0:
+            return self.get_manifold_size() * mean(self.grid_values)
         value_scale = backend_max(abs(self.grid_values))
         if not bool(isfinite(value_scale)) or bool(value_scale == 0.0):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing after the full matrix exposed JAX and PyTorch failures. The motivating case requires cancelling grid values near the floating-point maximum; repairing this across every backend would add disproportionate complexity for a low-priority edge case. NumPy, packaging, documentation, lint, security, and dependency checks passed.

PR #4458, opened 2026-07-16
Reject overflowing measurement variances

FlorianPfaff:
## Bug

`diagonal_measurement_covariance()` validated that supplied standard deviations were finite and positive, but squared them without validating the result. A finite standard deviation slightly larger than `sqrt(np.finfo(float).max)` therefore produced an infinite variance and an invalid covariance matrix.

The defect propagated through block-diagonal covariance construction and both weak-measurement model constructors.

## Fix

- square validated standard deviations under a local NumPy error state
- reject nonfinite variances with a clear `ValueError`
- preserve support for the boundary value whose variance is the largest representable finite float
- add focused regression coverage for the public helper, block helper, masked model, and weak-dimension model

## Validation

- reconstructed the source from the current Git blob and verified it matched blob `b38cad4a704f105bf2b286048b9c0f5ca6aaf37f` before editing
- syntax-compiled the modified implementation and regression test
- ran the focused regression in an isolated NumPy-backed package harness: `5 passed`
- branch is 2 commits ahead of `main`, 0 behind, and changes only the weak-measurement helper plus one focused regression file

Full repository CI is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/models/weak_measurement.py
+++ b/src/pyrecest/models/weak_measurement.py
@@ -32,7 +32,11 @@ _NON_REAL_NUMERIC_SCALAR_TYPES = (
 def diagonal_measurement_covariance(stds: Sequence[float] | np.ndarray) -> np.ndarray:
     "Return a diagonal covariance matrix from measurement standard deviations."
     values = _standard_deviations_array(stds)
-    return np.diag(values**2)
+    with np.errstate(over="ignore", invalid="ignore"):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded by #4460. That PR fixes the same overflow path, also rejects variance underflow to zero, and covers the same downstream measurement-model constructors. Keeping both would duplicate implementation and regression coverage.

PR #4465, opened 2026-07-16
Validate empty tail-selection controls

FlorianPfaff:
## Summary
- validate retention and quantile controls before returning from empty protected-tail selections
- validate rescue fractions for empty tail-rescue selections
- preserve the existing empty boolean-mask result for valid configurations
- add focused regressions for invalid controls on empty batches

## Bug
`protected_tail_topk_mask()` and `tail_rescue_topk_mask()` returned immediately when all three score vectors were empty. That bypassed the same scalar validation performed for nonempty batches, so malformed values such as a negative `retention_fraction`, `tail_quantile=0`, boolean `min_count`, or `rescue_fraction=0` were silently accepted only when a batch happened to be empty.

## Fix
Let the existing retained-count and quantile helpers validate their controls before the empty result is produced. For the rescue selector, compute the bounded rescue quota before the empty-tail fallback so `rescue_fraction` is validated consistently as well.

## Validation
- isolated regression harness covering valid and invalid empty-input paths passed
- diff is limited to `selection.py` and one focused regression file

Full repository tests were not runnable locally because this runtime has neither the GitHub CLI nor direct network checkout access; GitHub Actions should provide the repository-level validation.

### DIFF
--- a/src/pyrecest/evaluation/selection.py
+++ b/src/pyrecest/evaluation/selection.py
@@ -378,8 +378,6 @@ def protected_tail_topk_mask(
             "primary_scores, tail_scores, and reliability_scores must have the same length."
         )
     count = int(primary.shape[0])
-    if count == 0:
-        return np.zeros((0,), dtype=bool)
 
     retained_total = retained_count_from_fraction(
         count,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value relative to the current CI backlog. This only makes malformed controls fail consistently on an already-empty selection batch; it does not change valid selection results. The behavior can be folded into a broader validation cleanup later.

PR #4470, opened 2026-07-16
Reject masked random seeds

FlorianPfaff:
## Bug

`seed_all()` extracted scalar values with `.item()` before considering NumPy masked-array semantics. For masked inputs, `.item()` exposes the hidden payload instead of the missing value:

- `np.ma.masked` was silently normalized to seed `0`
- `np.ma.array(7, mask=True)` was silently normalized to seed `7`

Both inputs violate the public contract that a seed must be a non-negative integer or `None`, and they can make a missing configuration look reproducible while actually selecting an unintended seed.

## Fix

- reject scalar inputs for which `np.ma.is_masked()` is true before temporal, shape, and integer normalization
- retain support for unmasked integer-valued `MaskedArray` scalars
- add regressions for the masked constant, a masked scalar payload, and the unmasked scalar path

## Validation

- `python -m py_compile` passed for the modified module and test
- isolated `unittest` execution passed: 2 tests
- diff is limited to `src/pyrecest/reproducibility.py` and `tests/test_reproducibility_seed.py` with 3 additions in each file

The repository-level backend and CI matrices should run on this pull request.


FlorianPfaff:
Closing as low-value relative to the current CI backlog. Rejecting genuinely masked seed scalars is internally consistent, but this is an unusually narrow configuration edge case and can be included later in a consolidated scalar-validation change.

PR #4473, opened 2026-07-16
Freeze measurement score indices

FlorianPfaff:
## Bug

`MeasurementScore` is a frozen dataclass in a module explicitly providing non-mutating result containers, but it retained the caller's mutable `active_measurement_indices` list directly.

That allowed external mutation to change an already-created score. In particular, appending to an originally empty list could silently change `score.is_active` from `False` to `True`, potentially causing update logic to treat a skipped measurement batch as active.

## Fix

- accept sequence-like index inputs as before
- normalize `active_measurement_indices` to an owned tuple in `__post_init__`
- preserve NumPy index-array support and existing tracker iteration/length behavior
- add a regression that mutates the source list after construction and verifies the score remains inactive

## Validation

- syntax-compiled the modified module and test
- ran a focused runtime regression for caller-list mutation and NumPy array conversion
- diff is limited to the measurement-score container and its focused test

The full repository matrix should run on this pull request.

### DIFF
--- a/src/pyrecest/filters/measurement_scoring.py
+++ b/src/pyrecest/filters/measurement_scoring.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing after a second validation attempt. The branch remains red specifically across the NumPy 3.11, 3.12, and 3.14 matrix jobs, despite packaging, documentation, lint, security, JAX, and PyTorch paths being healthy. The change is a small defensive immutability hardening, and resolving the backend-specific compatibility fallout is disproportionate to its current value. The focused patch remains recoverable if this mutation pattern is observed in real use.

PR #4479, opened 2026-07-16
Reject complex rejection-sampler densities

FlorianPfaff:
## Bug

`FibonacciRejectionSampler._evaluate_pdf()` converted density outputs directly with `np.asarray(..., dtype=float)`. NumPy accepts native complex arrays by discarding their imaginary components with a warning.

A callable returning values such as `0.5 + 1.0j` could therefore be treated as the real density `0.5`, allowing invalid probability-density values to determine rejection decisions.

## Fix

- inspect the density output before float conversion
- reject native and object-wrapped complex values with a clear `ValueError`
- retain scalar broadcasting, shape validation, finiteness checks, and ordinary real-valued behavior
- install the validation through the sampling package's idempotent compatibility hook
- add a focused regression for complex rejection-density outputs and a real-valued control

## Impact

Invalid complex densities now fail fast instead of being silently projected onto their real components.

## Validation

- diff is limited to the sampling compatibility hook and one focused test module
- repository CI should exercise the new regression and the existing sampler suite

### DIFF
--- a/src/pyrecest/sampling/__init__.py
+++ b/src/pyrecest/sampling/__init__.py
@@ -1,3 +1,5 @@
+import numpy as _np
+
 from .abstract_sampler import AbstractSampler
 from .euclidean_sampler import (
     AbstractEuclideanSampler,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded by #4481, which fixes `FibonacciRejectionSampler._evaluate_pdf()` in its defining module. This avoids an import-path-dependent monkeypatch and covers direct submodule imports.

PR #4480, opened 2026-07-16
Reject masked evidence boolean flags

FlorianPfaff:
## Bug

`_coerce_bool_flag()` converted inputs with `np.asarray()` before accounting for NumPy masked-array semantics. For a masked Boolean scalar, `np.asarray()` exposes the hidden payload instead of the missing value.

Consequently, values such as `np.ma.array(True, mask=True)` and `np.ma.array(False, mask=True)` were silently accepted as evidence-mode controls. This could select full smoothing or evidence-only execution from a missing configuration value rather than rejecting it.

## Fix

- reject genuinely masked Boolean inputs before array conversion
- apply the validation consistently to `return_smoothed` and `terminal_posterior` construction paths
- preserve support for `MaskedArray` Boolean scalars whose mask is false
- extend the existing Boolean-conversion regression suite across all public evidence-mode factories

## Validation

- reproduced the old behavior: masked `True` and masked `False` were converted to ordinary booleans
- isolated regression passed for masked rejection and unmasked compatibility
- diff is limited to two source lines and focused additions to the existing evidence Boolean test module

The repository CI matrix should provide full integration validation.

### DIFF
--- a/src/pyrecest/evidence.py
+++ b/src/pyrecest/evidence.py
@@ -28,6 +28,8 @@ _RESERVED_DIAGNOSTIC_METADATA_KEYS = frozenset(
 def _coerce_bool_flag(value: bool, name: str) -> bool:
     """Return a bool flag without treating arbitrary truthy objects as true."""
 
+    if np.ma.is_masked(value):
+        raise ValueError(f"{name} must be a bool")
     try:
         value_array = np.asarray(value)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value relative to the current CI backlog. This only rejects masked Boolean scalars for two evidence-mode factory flags; it is better folded into a consolidated masked-control validation change than run through the full matrix independently.

PR #4483, opened 2026-07-17
Reject masked HistoryRecorder boolean flags

FlorianPfaff:
## Bug

`HistoryRecorder._validate_bool_flag()` converted flag inputs with the active backend before accounting for NumPy masked-array semantics. For a genuinely masked Boolean scalar, conversion exposes the hidden payload rather than the missing value.

Consequently, values such as `np.ma.array(True, mask=True)` and `np.ma.array(False, mask=True)` were silently accepted for `pad_with_nan` and `copy_value`. A missing configuration value could therefore enable or disable padded storage or deep-copy behavior based on an inaccessible payload.

## Fix

- reject genuinely masked Boolean controls before backend conversion
- apply the validation consistently to `register(..., pad_with_nan=...)`, `record(..., pad_with_nan=...)`, and `record(..., copy_value=...)`
- preserve Python, NumPy, backend scalar, and unmasked `MaskedArray` Boolean support
- add focused regressions for both hidden Boolean payloads and all public flag paths

## Validation

- reproduced the old NumPy behavior: `np.asarray()` exposes the hidden Boolean payload
- isolated validation passed for masked rejection and ordinary/unmasked Boolean compatibility
- branch comparison against `main`: 2 files, +42/-0, 2 commits ahead, 0 behind

The full repository backend and CI matrices should provide integration validation.

### DIFF
--- a/src/pyrecest/utils/history_recorder.py
+++ b/src/pyrecest/utils/history_recorder.py
@@ -22,6 +22,8 @@ class _HistoryEntry:
 
 
 def _validate_bool_flag(value: Any, name: str) -> bool:
+    if np.ma.is_masked(value):
+        raise TypeError(f"{name} must be a boolean")
     if isinstance(value, bool):
         return value
     try:
--- a/tests/test_history_recorder_masked_flags.py
+++ b/tests/test_history_recorder_masked_flags.py
@@ -1,0 +1,40 @@
+import unittest
+
+import numpy as np
+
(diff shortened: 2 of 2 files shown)


FlorianPfaff:
Closing as low-value relative to the current CI backlog. Masked Boolean handling in `HistoryRecorder` is a narrow defensive edge case and should be bundled with a broader recorder/control-validation cleanup rather than consume an independent full backend matrix.

PR #4488, opened 2026-07-17
Reject masked sample counts in Gaussian and Dirac distributions

FlorianPfaff:
## Bug

NumPy masked scalars expose their hidden payload through ordinary scalar conversion:

```python
np.asarray(np.ma.array(3, mask=True)).item() == 3
operator.index(np.ma.array(3, mask=True)) == 3
```

As a result, both `GaussianDistribution.sample(...)` and Dirac-distribution sampling accepted a genuinely masked count and silently drew the hidden number of samples. A missing configuration value could therefore trigger real work instead of failing validation.

## Fix

- reject NumPy masked scalar counts before payload-exposing conversion
- apply the same contract to Gaussian and Dirac sample-count validators
- continue accepting integer `MaskedArray` scalars whose mask is explicitly false

## Validation

Added focused regression coverage for:

- `np.ma.masked`
- `np.ma.array(3, mask=True)`
- `np.ma.array(2, mask=False)` preserving valid behavior

The final diff is limited to two validation guards and the regression test file.

### DIFF
--- a/src/pyrecest/distributions/abstract_dirac_distribution.py
+++ b/src/pyrecest/distributions/abstract_dirac_distribution.py
@@ -36,6 +36,8 @@ from .abstract_distribution_type import AbstractDistributionType
 def _validate_positive_sample_count(n) -> int:
     """Return ``n`` as a positive Python int after scalar-count validation."""
     message = "n must be a positive integer."
+    if np.ma.isMaskedArray(n) and bool(np.ma.getmaskarray(n).any()):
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing as low-value relative to the current CI backlog. Masked sample-count handling is a narrow defensive edge case and should be bundled into a consolidated distribution input-validation change rather than run through a separate full backend matrix.

PR #4489, opened 2026-07-17
Reject masked tracking event inputs

FlorianPfaff:
## Bug

`tracking.event_records` converted inputs with `np.asarray()` before accounting for NumPy masked-array semantics. For genuinely masked values, NumPy exposes the hidden payload rather than preserving the missing value.

Consequently, inputs such as a masked timestamp, measurement entry, covariance entry, NIS value, or accepted flag were silently interpreted from their inaccessible payload. A missing tracking value could therefore enter a replay record as valid numeric data or select an update decision.

## Fix

- reject genuinely masked values in the shared real-numeric validation path
- reject masked scalar values before finite-scalar conversion
- reject masked Boolean values before accepted-flag conversion
- retain support for `MaskedArray` inputs whose mask is entirely false

## Regression coverage

- masked `TrackingEvent` time, measurement, covariance, and accepted inputs
- masked `TrackingRecord` state, covariance, innovation, NIS, and accepted inputs
- compatibility coverage for unmasked `MaskedArray` scalars, vectors, and matrices

## Validation

- reproduced NumPy exposing hidden payloads through `np.asarray()`
- branch comparison against `main`: 2 files, +138/-0, 2 commits ahead, 0 behind
- source diff is limited to six validation lines; the full backend and repository matrix is delegated to GitHub Actions

### DIFF
--- a/src/pyrecest/tracking/event_records.py
+++ b/src/pyrecest/tracking/event_records.py
@@ -45,6 +45,8 @@ def _dtype_is_temporal(dtype: np.dtype) -> bool:
 
 
 def _raw_item_is_invalid_numeric(value: Any) -> bool:
+    if np.ma.is_masked(value):
+        return True
     if isinstance(value, _INVALID_NUMERIC_SCALAR_TYPES):
         return True
     if isinstance(value, np.ndarray):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value relative to the current CI backlog. This is another broad masked-array defensive check with a large regression-only diff; it should be consolidated with the other masked tracking-input validations rather than consume an independent full matrix.

PR #4568, opened 2026-07-22
Stabilize Merwe sigma-point scaling

FlorianPfaff:
## Bug

`MerweScaledSigmaPoints` reconstructed its positive sigma-point scale through

```python
lam = alpha**2 * (n + kappa) - n
scale = n + lam
```

For sufficiently small but representable positive `alpha`, the subtraction rounds `lam` to exactly `-n`; adding `n` back then produces zero. For example, `n=2`, `alpha=1e-10`, and `kappa=0` have the valid direct scale `2e-20`, while the cancellation-prone reconstruction yields `0.0`. Weight construction then divides by zero and sigma-point generation uses a degenerate Cholesky scale.

## Fix

- compute `alpha**2 * (n + kappa)` directly once during construction;
- require the resulting scale to be finite and strictly positive;
- reuse that direct scale for both weight construction and covariance scaling;
- reject scales that underflow to zero or overflow to infinity with a clear `ValueError`.

## Impact

Valid tightly clustered Merwe sigma-point sets no longer fail or collapse because of avoidable cancellation. Ordinary parameter choices retain their existing weights and moment-matching behavior.

## Regression coverage

- verify `alpha=1e-10` produces the expected nonzero sigma-point displacement;
- verify all weights remain finite;
- reject underflowing and overflowing scale configurations explicitly.

## Validation

- isolated NumPy regression harness passed for the small-alpha case, ordinary moment matching, and scale-boundary rejection;
- both changed Python files pass syntax compilation;
- branch is 2 commits ahead of current `main`, 0 behind;
- diff is limited to the sigma-point implementation and one focused test module.

GitHub Actions should provide the authoritative repository-wide and backend-matrix validation.

This revisits the still-applicable defect previously identified in closed, unmerged PR #4447, verified against current `main`.


FlorianPfaff:
Closing after CI triage. The proposed `alpha=1e-10` support produces float64 Merwe weights of order 1e20 whose normalization depends on an unrepresentable unit-scale cancellation; the NumPy matrix fails consistently while JAX/PyTorch behavior differs. A correct cross-backend solution needs a redesigned or explicitly bounded weighting contract, not a focused patch to the scale expression.

PR #4574, opened 2026-07-22
Reject complex LTI discretization inputs

FlorianPfaff:
## Bug

`continuous_to_discrete_lti()` converted its system and noise matrices with `np.asarray(..., dtype=float)`. NumPy accepts complex arrays on that path by discarding their imaginary components with a warning, so the function could discretize a different real-valued model instead of rejecting an unsupported complex model.

The issue affected `continuous_matrix`, `noise_input_matrix`, and `continuous_noise_covariance`, including object arrays containing complex scalars.

## Fix

- validate all three matrix inputs before the existing float conversion
- reject native complex arrays and object arrays containing complex scalars with a parameter-specific `ValueError`
- preserve the existing real-valued discretization implementation and public import path
- add focused regression coverage for every affected parameter and both complex representations

## Impact

Callers can no longer receive plausible-looking transition or process-noise matrices after imaginary components were silently removed.

## Validation

- Python syntax parsing passed for the changed implementation and regression
- isolated reproduction confirmed the old NumPy conversion truncates the imaginary part
- isolated checks confirmed the new validation rejects native and object-wrapped complex matrices while accepting real matrices

GitHub Actions should provide the authoritative repository-wide validation.

### DIFF
--- a/src/pyrecest/models/_validated_motion_models.py
+++ b/src/pyrecest/models/_validated_motion_models.py
@@ -4,9 +4,49 @@ from __future__ import annotations
 
 from typing import Any
 
+import numpy as np
+
 from . import motion_models as _motion_models
 
+_continuous_to_discrete_lti_impl = _motion_models.continuous_to_discrete_lti
 _nearly_coordinated_turn_model_impl = _motion_models.nearly_coordinated_turn_model
+
+
+def _reject_complex_matrix(value: Any, name: str) -> None:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant with #4576, which addresses the same complex-valued LTI discretization defect on current `main` with the same parameter coverage and a cleaner validation-wrapper integration. Keeping one implementation avoids duplicate CI and review work.

PR #4624, opened 2026-07-23
Reject invalid Python sequence indices in PyTorch take

FlorianPfaff:
## Bug

The PyTorch `take()` compatibility contract validates tensors and NumPy arrays, but plain Python lists and tuples bypass validation. Fractional, complex, or textual index sequences can therefore reach backend coercion paths instead of being rejected consistently with NumPy array inputs.

For example, `[0.0, 1.0]` was not checked even though `np.array([0.0, 1.0])` was rejected.

## Fix

- normalize non-tensor index inputs with `np.asarray` for dtype inspection;
- apply the existing `same_kind` castability check to all non-scalar array-like inputs;
- preserve valid integer and Boolean sequences;
- preserve scalar handling in the underlying implementation.

## Regression coverage

Focused tests cover plain Python lists and tuples containing floating-point, complex, and textual indices, plus valid integer/Boolean sequences and scalar passthrough.

## Scope

The branch is based on current `main`, is 2 commits ahead and 0 behind, and changes only the take-index compatibility validator plus focused tests.

GitHub Actions provides the authoritative repository-wide validation.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_take_index_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_take_index_contract.py
@@ -17,8 +17,13 @@ def _validate_take_indices(indices, torch_module):
             raise TypeError(_TAKE_INDEX_TYPE_MESSAGE)
         return indices
 
-    if isinstance(indices, np.ndarray) and not np.can_cast(
-        indices.dtype,
+    try:
+        indices_array = np.asarray(indices)
+    except (TypeError, ValueError):
+        return indices
+
+    if indices_array.ndim > 0 and not np.can_cast(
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value/incompatible with the repository's established backend contract. The existing portable test explicitly requires Python floating-point index lists such as `[1.5]` to retain NumPy-compatible coercion behavior, while typed floating-point arrays remain rejected. This PR changes that intentional distinction and therefore fails the PyTorch matrix. Resolving it would require a broader, explicitly approved API-contract change rather than a focused bug fix.

PR #4651, opened 2026-07-24
Reject non-integer reproducibility seeds

FlorianPfaff:
## Bug

`seed_all()` and `temporary_seed()` document their seed as an integer, but `_normalize_seed()` converted any scalar that compared equal to an integer. As a result, values such as `1.0`, `Decimal("1")`, and `Fraction(1, 1)` were silently accepted.

This weakens the public type contract and can hide configuration or serialization errors where a seed was supplied with the wrong scalar type.

## Fix

- require the normalized scalar to implement `numbers.Integral`;
- continue accepting Python and NumPy integer scalars;
- preserve the existing non-negative and 32-bit backend range checks.

## Regression coverage

A focused test verifies rejection of integer-valued floats, `Decimal`, and `Fraction` seeds, while confirming that Python and NumPy integer scalar types remain accepted.

## Validation

The branch is based on the current `main` head and is two commits ahead, zero behind. The compare diff is limited to the seed validator and one focused regression test. GitHub Actions will provide the authoritative backend-matrix validation.

### DIFF
--- a/src/pyrecest/reproducibility.py
+++ b/src/pyrecest/reproducibility.py
@@ -69,22 +69,12 @@ def _normalize_seed(seed: int | None) -> int | None:
     except (TypeError, ValueError, RuntimeError, OverflowError) as exc:
         raise ValueError(message) from exc
 
-    if isinstance(scalar, _UNSUPPORTED_SEED_SCALAR_TYPES):
+    if isinstance(scalar, _UNSUPPORTED_SEED_SCALAR_TYPES) or not isinstance(
+        scalar, Integral
+    ):
         raise ValueError(message)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value regression risk. This is a deliberate public-contract tightening rather than a correctness fix: the existing normalizer intentionally accepts scalar values that are exactly equal to an integer, while this branch rejects float, Decimal, and Fraction representations. The result is failure across the NumPy, JAX, and PyTorch test matrix on supported Python versions, despite lint, docs, packaging, and security checks passing. Preserving the established compatibility is preferable to repairing the suite around this stricter policy.

PR #4677, opened 2026-07-24
Validate extreme Merwe sigma-point scaling

FlorianPfaff:
## Summary

- compute the Merwe scaling factor without the cancellation-prone `lambda` round trip
- reject underflowed, overflowed, or otherwise non-finite scaling factors with a clear `ValueError`
- validate that the resulting covariance and off-center weights are finite
- add regressions for finite `alpha` values that previously raised accidental `ZeroDivisionError` or `OverflowError`

## Root cause

The implementation computed

```python
lambda_ = alpha**2 * (n + kappa) - n
scale = n + lambda_
```

For very small finite `alpha`, `alpha**2` underflowed and the subtraction/addition sequence collapsed `scale` to zero. For very large finite `alpha`, exponentiation raised `OverflowError`. Both bypassed the constructor's intended validation contract.

## Validation

- reproduced the old failures for `alpha=1e-300` and `alpha=1e308`
- added `tests/test_sigma_points_extreme_scale.py`
- reviewed the final diff for NumPy/JAX/PyTorch backend portability

GitHub Actions will provide the full repository test matrix.

### DIFF
--- a/src/pyrecest/sampling/sigma_points.py
+++ b/src/pyrecest/sampling/sigma_points.py
@@ -169,13 +169,10 @@ class MerweScaledSigmaPoints:
         central_covariance_weight = central_mean_weight + (
             1.0 - alpha_squared + self.beta
         )
-        if not all(
-            math.isfinite(weight)
-            for weight in (
-                central_mean_weight,
-                central_covariance_weight,
-                off_center_weight,
-            )
+        if not (
+            math.isfinite(central_mean_weight)
+            and math.isfinite(central_covariance_weight)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing during PR cleanup: the proposed extreme-scale guard is low practical value (it targets pathological `alpha` values such as `1e-300` and `1e308`) and the current implementation fails the full NumPy matrix while passing JAX/PyTorch. Repairing it safely would require a broader backend-numerics redesign and regression audit, so it is not worth carrying as an open PR.

PR #4685, opened 2026-07-24
Reject nested invalid assignment cost values

FlorianPfaff:
## Bug

Assignment cost validation inspected only the immediate elements of object arrays. A zero-dimensional complex array nested inside a cost matrix or non-assignment cost vector therefore bypassed the complex-value check.

The later floating-point conversion emitted a `ComplexWarning` and silently discarded the imaginary component, so Murty and maximum-cardinality assignment could solve a different real-valued problem than the caller supplied.

## Fix

- recursively inspect NumPy arrays, lists, tuples, and backend-convertible arrays for invalid scalar types;
- guard recursive inspection against cyclic containers;
- reject nested complex cost-matrix and non-assignment values before any floating-point cast;
- preserve support for nested zero-dimensional real scalar arrays.

## Validation

- reproduced the old silent complex-to-real cast for cost matrices and row/column non-assignment costs;
- `python -m py_compile` passed for the modified module and regression test;
- isolated NumPy/SciPy execution passed all three new regression tests;
- additional smoke checks covered nested Boolean, text, temporal, complex, real, and cyclic object values.

GitHub Actions should run the full repository and backend matrices.

### DIFF
--- a/src/pyrecest/utils/assignment.py
+++ b/src/pyrecest/utils/assignment.py
@@ -107,44 +107,61 @@ def _has_temporal_dtype(value) -> bool:
         return "datetime64" in dtype_name or "timedelta64" in dtype_name
 
 
-def _contains_boolean_values(value) -> bool:
-    if isinstance(value, _BOOLEAN_TYPES):
+def _contains_values_of_type(
+    value,
+    types: tuple[type, ...],
+    active_ids: set[int] | None = None,
+) -> bool:
+    if isinstance(value, types):
         return True
-    try:
-        values = _np.asarray(_to_numpy(value), dtype=object).reshape(-1)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing during PR cleanup. The recursive validator is useful in principle, but the implementation now descends into ordinary NumPy scalar leaves and calls the active backend's `to_numpy` on them. That breaks the PyTorch matrix (at least Python 3.12 and 3.13) while NumPy/JAX pass. A safe replacement should narrowly recurse only through containers/backend arrays and stop at allowed scalar leaves; because this helper is shared by Boolean, text, temporal, and complex validation, repairing it here would require a broader cross-backend regression pass rather than a low-risk patch.

PR #4699, opened 2026-07-24
Fix squeeze validation for non-singleton axes

FlorianPfaff:
## Summary

- reject explicit NumPy/Autograd squeeze axes whose extent is not one
- preserve existing singleton-axis and out-of-bounds behavior
- add focused regression coverage for the public backend facade

## Bug

`backend.squeeze(values, axis=...)` silently returned the unchanged array when the requested axis had size greater than one. NumPy raises `ValueError` for this invalid operation. The silent return can hide incorrect shape assumptions and allow downstream code to continue with an unexpected rank.

## Validation

- added regression tests for non-singleton, singleton, and out-of-bounds axes
- full validation is delegated to GitHub Actions because the local execution environment cannot resolve GitHub to install or run the checkout

### DIFF
--- a/src/pyrecest/backend_support/_shared_numpy_squeeze_contract.py
+++ b/src/pyrecest/backend_support/_shared_numpy_squeeze_contract.py
@@ -1,0 +1,40 @@
+"""Compatibility patch for NumPy-style squeeze axis validation."""
+
+from __future__ import annotations
+
+
+def patch_shared_numpy_squeeze_nonsingleton_axis_contract() -> None:
+    """Reject explicitly selected axes whose extent is not one."""
+    try:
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing because this changes an established, explicitly tested PyRecEst backend contract rather than fixing an isolated defect. `tests/test_backend_arraylike_contracts.py::test_squeeze_non_singleton_axis_is_noop` requires non-singleton `squeeze(axis=...)` to remain a no-op across backends. The runtime patch also imports the shared NumPy backend before checking the active backend, causing failures in the JAX matrix. Adopting NumPy's exception semantics would need an intentional cross-backend API change with migration and coordinated test updates, not a compatibility patch.

PR #4707, opened 2026-07-25
Reject fractionally rounded SE2 Bingham sample counts

FlorianPfaff:
## Bug

`SE2BinghamDistribution.sample` validates `n` by converting the original scalar to binary64 and calling `float.is_integer()`. Above binary64's consecutive-integer range, an exact fractional value can lose its fractional part during conversion.

For example, `Fraction(2**54 + 1, 2)` is exactly a half-integer, but converts to the integer-valued float `2**53`. The validator therefore accepts it and can proceed toward an unintended enormous allocation.

## Fix

- require exact equality between the original scalar and its integer conversion;
- preserve rejection of booleans, non-scalars, fractional values, non-finite values, and non-positive counts;
- preserve exact large integer-valued inputs;
- install the corrected validator when the Cartesian-product distribution package is initialized.

## Regression coverage

- rejects the exact half-integer whose fractional part disappears in binary64;
- accepts the neighboring exact large integer without allocating samples.

GitHub Actions should provide the authoritative full backend and repository validation.

### DIFF
--- a/src/pyrecest/distributions/cart_prod/__init__.py
+++ b/src/pyrecest/distributions/cart_prod/__init__.py
@@ -1,0 +1,33 @@
+"""Cartesian-product distributions and compatibility fixes."""
+
+from __future__ import annotations
+
+from . import se2_bingham_distribution as _se2_bingham_distribution
+
+
+def _validate_se2_bingham_sample_count(n) -> int:
+    """Validate SE(2) Bingham sample counts without binary64 rounding."""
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value relative to repair cost: this narrow binary64 half-integer edge-case PR is conflicted with the newly merged validation work and currently fails both the main test workflow and documentation examples. The broader exact-integer validation fixes already merged in #4704 and #4706 cover the same failure pattern in higher-value public paths.

PR #4708, opened 2026-07-25
Fix exact multi-session integer validation

FlorianPfaff:
## Summary

- validate `max_gap` and index-matrix `fill_value` against the original numeric scalar instead of a binary64-rounded copy;
- reject exact fractional values that happen to round to integer-valued floats;
- preserve support for ordinary integers, integral floats, and exact rational integers;
- add focused regressions for both validation paths.

## Bug

The multi-session score helpers converted non-integer numeric scalars to `float` and then called `is_integer()`. Above binary64's consecutive-integer range, an exact fractional value can therefore be accepted as an integer. For example, `Fraction(2**54 + 1, 2)` is exactly `9007199254740992.5`, but converts to `9007199254740992.0`.

For `max_gap`, accepting such a value can silently alter which inter-session edges are admitted to the assignment graph. For `fill_value`, the same rounding can silently replace an invalid fractional sentinel with an integer sentinel.

## Fix

A shared helper now converts the original scalar to `int` and requires exact equality with that original scalar. The existing sign constraints remain unchanged: `max_gap` must be non-negative and `fill_value` must be negative.

## Validation

- reproduced the pre-fix binary64 rounding behavior for both positive and negative exact fractions;
- added regression coverage that rejects the rounded fractions and accepts neighboring exact rational integers;
- verified the branch is two commits ahead and zero behind `main`, changing only the validator module and its focused test file;
- full repository CI is delegated to GitHub Actions because this execution environment has no working GitHub CLI or networked checkout.

### DIFF
--- a/src/pyrecest/utils/multisession_assignment_score.py
+++ b/src/pyrecest/utils/multisession_assignment_score.py
@@ -144,55 +144,42 @@ def _normalize_min_score(min_score: Any) -> float:
     return min_score_float
 
 
-def _normalize_max_gap(max_gap: Any) -> int:
-    max_gap_array = np.asarray(max_gap)
-    if max_gap_array.shape != () or _is_rejected_scalar_array(max_gap_array):
-        raise ValueError("max_gap must be a non-negative integer.")
+def _normalize_exact_integer(value: Any, message: str) -> int:
+    """Return an exact integer scalar without binary64 rounding."""
+    value_array = np.asarray(value)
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value and redundant with the exact-integer validation work already merged. This 106-line two-path change targets only exact fractional values above binary64's consecutive-integer range; the practical benefit does not justify another specialized helper and regression surface.

PR #4711, opened 2026-07-25
Reject fractionally rounded piecewise sample counts

FlorianPfaff:
## Bug

`_validate_positive_sample_count` converted non-integer real values to binary64 before checking `is_integer()`. Exact fractions beyond binary64's consecutive-integer range can therefore round to an integer-valued float. For example, `Fraction(2**54 + 1, 2)` was accepted as `9007199254740992`, potentially causing an enormous sample allocation or interval loop.

## Fix

- compare the original scalar with its integer conversion;
- reject values that are not exactly integral;
- preserve exact rational integers and the existing supported numeric scalar forms.

## Regression coverage

- rejects an exact fraction that binary64 rounds to an integer;
- accepts a neighboring exact rational integer.

## Validation

- reproduced the previous binary64 rounding behavior independently;
- exercised the fixed validation logic for the fractional reproducer, an exact rational integer, an integral NumPy float, and the existing invalid-value cases;
- branch is two commits ahead and zero behind `main`, changing only the validator and its focused test file.

Full repository pytest was not available because this environment has no GitHub CLI and cannot clone through outbound Git. GitHub Actions should provide the authoritative full-suite validation.

### DIFF
--- a/src/pyrecest/distributions/circle/piecewise_constant_distribution.py
+++ b/src/pyrecest/distributions/circle/piecewise_constant_distribution.py
@@ -51,11 +51,11 @@ def _validate_positive_sample_count(n) -> int:
 
     try:
         count_int = int(count)
-        count_float = float(count)
+        is_exact_integer = bool(count == count_int)
     except (OverflowError, TypeError, ValueError) as exc:
         raise ValueError("n must be an integer") from exc
 
-    if not np.isfinite(count_float) or not count_float.is_integer():
+    if not is_exact_integer:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low-value redundancy. This addresses the same pathological exact-Fraction-above-binary64 edge case as the recently merged integer-validation PRs, but in another narrow sampling helper. The maintenance and regression surface is not justified by the practical likelihood of this input.

PR #4741, opened 2026-07-25
Reject non-finite hypertoroidal inputs

FlorianPfaff:
## Bug

The shared hypertorus input-normalization helpers accepted `NaN`, `+inf`, and `-inf` as valid real angles. These values then reached wrapping, density, and shifting operations, where they could silently propagate non-finite outputs through otherwise valid distribution objects.

## Fix

- validate backend arrays with the active backend's `isfinite`/`all` operations;
- reject non-finite shift vectors and hypertoroidal evaluation points at the shared input boundary;
- preserve the existing boolean, complex, scalar, and shape validation behavior.

## Regression coverage

Added focused tests for `NaN`, positive infinity, and negative infinity through both `as_shift_vector` and `as_hypertoroidal_points`, plus a valid-input shape check.

## Scope

The branch is two commits ahead and zero behind `main`. The production change is limited to 13 additions and one deletion in the shared hypertorus input validator.

### DIFF
--- a/src/pyrecest/distributions/hypertorus/_input_validation.py
+++ b/src/pyrecest/distributions/hypertorus/_input_validation.py
@@ -3,7 +3,7 @@
 import numpy as np
 
 # pylint: disable=no-name-in-module,no-member
-from pyrecest.backend import array
+from pyrecest.backend import all, array, isfinite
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4747, which combines this non-finite-angle validation with the overlapping temporal-angle validation on current main and retains both regression suites.

PR #4743, opened 2026-07-25
Reject temporal hypertoroidal angle inputs

FlorianPfaff:
## Summary

- reject NumPy and Python datetime/duration values in shared hypertoroidal angle normalization
- apply the validation to both shift vectors and evaluation points before and after backend conversion
- add regression coverage for native temporal scalars, Python temporal objects, object-wrapped values, and the public circular-uniform shift path

## Bug

The shared hypertoroidal input helpers rejected boolean and complex values but did not reject temporal values. Nanosecond-resolution `numpy.datetime64` and `numpy.timedelta64` scalars can expose integer payloads during conversion, while `CircularUniformDistribution.shift(...)` ignores the shift entirely because the distribution is invariant.

Consequently, malformed temporal inputs such as `np.timedelta64(1, "ns")` could be silently accepted as angular shifts instead of being rejected. This can hide unit/type errors at API boundaries.

## Fix

Add explicit temporal-value detection for:

- NumPy `datetime64` and `timedelta64`
- Python `date`, `datetime`, and `timedelta`
- temporal values wrapped in object arrays

The checks run before backend conversion and again afterward, matching the existing boolean and complex validation pattern.

## Validation

- source and regression test files pass `python -m py_compile`
- focused local validation confirms all temporal cases are rejected while ordinary numeric vectors remain accepted
- branch comparison: 2 commits ahead, 0 behind `main`
- diff is limited to 29 source additions and one focused 36-line regression test module

GitHub Actions should provide the authoritative full backend-matrix validation.

### DIFF
--- a/src/pyrecest/distributions/hypertorus/_input_validation.py
+++ b/src/pyrecest/distributions/hypertorus/_input_validation.py
@@ -1,5 +1,7 @@
 """Input-normalization helpers for hypertoroidal distributions."""
 
+import datetime as _datetime
+
 import numpy as np
 
 # pylint: disable=no-name-in-module,no-member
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Superseded by #4747, which combines this temporal-angle validation with the overlapping non-finite-angle validation on current main and retains both regression suites.

PR #4744, opened 2026-07-25
Reject matrix-shaped UKF vectors

FlorianPfaff:
## Summary

- validate UKF transition outputs, measurement-function outputs, and observations as scalars or one-dimensional vectors
- reject row and column matrices instead of silently flattening them
- preserve scalar support for one-dimensional state and measurement models
- add focused regression coverage for all three vector boundaries

## Bug

The UKF core normalized transition outputs, measurement-function outputs, and observations with `reshape(..., (-1,))`. A malformed row or column matrix was therefore silently accepted whenever its element count matched the expected dimension.

This can conceal upstream shape errors and reinterpret matrix storage order as state or measurement component order, allowing an unintended model output or observation to enter the filter update.

## Fix

Introduce a shared backend-portable `_as_vector` coercion helper that:

- promotes scalars to one-element vectors;
- accepts one-dimensional vectors unchanged;
- rejects arrays with two or more dimensions with a clear `ValueError`.

## Regression coverage

The new tests verify that both row and column matrices are rejected for:

- nonlinear transition-function outputs;
- nonlinear measurement-function outputs;
- measurement observations.

A one-dimensional predict/update test confirms that scalar transition outputs, scalar measurement outputs, and scalar observations remain supported.

## Validation

- both changed Python files pass `py_compile` and AST parsing
- the committed comparison is limited to 15 additions/3 deletions in `_ukf.py` plus one focused 99-line regression module
- the two newer `main` commits touch unrelated replay-scoring and weak-measurement files, so the branch has no known overlap
- full backend and repository validation is delegated to GitHub Actions because this runtime cannot clone the repository or install its test environment

### DIFF
--- a/src/pyrecest/filters/_ukf.py
+++ b/src/pyrecest/filters/_ukf.py
@@ -35,6 +35,18 @@ __all__ = [
 ]
 
 
+def _as_vector(value, name):
+    """Coerce a scalar or vector without silently flattening matrices."""
+    value = asarray(value, dtype=float64)
+    if len(value.shape) == 0:
+        return reshape(value, (1,))
+    if len(value.shape) != 1:
+        raise ValueError(
+            f"{name} must be a scalar or one-dimensional vector; got shape {value.shape}"
+        )
+    return value
+
+
 # ---------------------------------------------------------------------------
 # Unscented Kalman Filter
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant with #4745. Both implement the same UKF scalar-or-vector boundary validation; #4745 retains the same production coverage and adds explicit state-preservation checks after rejected prediction outputs.

PR #4752, opened 2026-07-26
Reject matrix-shaped bias timestamp vectors

FlorianPfaff:
## Summary

- require bias-calibration timestamp inputs to be scalars or one-dimensional vectors
- reject row and column matrices instead of silently flattening them
- preserve scalar timestamp support for one-sample calibration
- add focused regression coverage for measurement and reference timestamps

## Bug

`make_bias_training_examples` passed `measurement_times_s` and `reference_times_s` through `_as_numeric_vector`, which unconditionally called `reshape(-1)`. As a result, matrix-shaped timestamps were silently flattened and accepted whenever their element count matched the corresponding measurement rows.

For example, a `1 x 2` measurement-time matrix was interpreted as a two-sample vector. This hides upstream shape errors and can change timestamp ordering without any validation failure.

## Fix

The vector coercion helper now:

- converts scalar numeric inputs to length-one vectors;
- accepts explicitly one-dimensional numeric arrays;
- raises `ValueError` for arrays with more than one dimension.

## Validation

- reproduced the pre-fix acceptance of row and column timestamp matrices
- added regressions covering both matrix orientations for measurement and reference timestamps
- verified scalar timestamp compatibility
- ran the focused regression tests locally: `5 passed`
- compiled the modified production and test modules successfully
- confirmed the branch is two commits ahead and zero behind `main`

### DIFF
--- a/src/pyrecest/calibration/bias.py
+++ b/src/pyrecest/calibration/bias.py
@@ -367,7 +367,12 @@ def _as_numeric_array(values: Any, name: str) -> np.ndarray:
 
 
 def _as_numeric_vector(values: Any, name: str) -> np.ndarray:
-    return _as_numeric_array(values, name).reshape(-1)
+    out = _as_numeric_array(values, name)
+    if out.ndim == 0:
+        return out.reshape(1)
+    if out.ndim != 1:
+        raise ValueError(f"{name} must be one-dimensional")
+    return out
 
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as a duplicate subset of #4751. PR #4751 includes the same bias timestamp-vector validation and also covers the related calibration time-offset helpers.

PR #4753, opened 2026-07-26
Reject matrix-shaped bias calibration timestamps

FlorianPfaff:
## Bug

`make_bias_training_examples` normalized timestamp inputs with `.reshape(-1)`. As a result, row and column matrices such as `(1, n)` and `(n, 1)` were silently flattened and accepted as timestamp vectors. This can conceal an upstream orientation error and pair measurements with timestamps under an unintended shape contract.

## Fix

- preserve scalar timestamps for the single-sample case;
- accept ordinary one-dimensional timestamp vectors;
- reject higher-dimensional `measurement_times_s` and `reference_times_s` with a clear `ValueError`;
- add regressions for row and column matrices on both timestamp inputs.

## Validation

- Before the fix: the four matrix-shape regression cases fail because no exception is raised.
- After the fix: `5 passed` in `tests/calibration/test_bias_time_vector_shapes.py`.
- `python -m py_compile src/pyrecest/calibration/bias.py` passes.

### DIFF
--- a/src/pyrecest/calibration/bias.py
+++ b/src/pyrecest/calibration/bias.py
@@ -367,7 +367,12 @@ def _as_numeric_array(values: Any, name: str) -> np.ndarray:
 
 
 def _as_numeric_vector(values: Any, name: str) -> np.ndarray:
-    return _as_numeric_array(values, name).reshape(-1)
+    out = _as_numeric_array(values, name)
+    if out.ndim == 0:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as a duplicate subset of #4751. The production change is equivalent to #4752, while #4751 retains the same validation and broader calibration coverage.

PR #4780, opened 2026-07-26
Make nonlinear particle predictions transactional

FlorianPfaff:
## Summary

- reject nonlinear transforms that change the particle-array shape
- keep nonvectorized predictions in a copied filter state until all additive-noise samples succeed
- commit vectorized predictions only after transform and noise-output validation
- add regressions for silent shape corruption and mid-prediction noise failure

## Bug

`AbstractParticleFilter.predict_nonlinear` handled its two execution modes inconsistently. The vectorized path assigned `f(filter_state.d)` without validating its shape, so a transform returning one scalar per multidimensional particle could silently corrupt the filter state. The nonvectorized path assigned the transformed filter state before sampling additive noise; if a later noise shift or sample raised, the filter remained partially predicted.

## Fix

Vectorized output is converted to a backend array and checked against the original particle shape before assignment. Nonvectorized output remains in the copied state returned by `apply_function`. Additive-noise samples are generated and shape-checked completely before either path commits its result, so failures leave the original particles and weights unchanged. The vectorized path retains its existing allocation-efficient behavior.

## Validation

- added focused regression coverage in `tests/filters/test_particle_filter_prediction_atomicity.py`
- verified the branch is based on current `main` and changes only the particle-filter implementation plus the focused test module
- GitHub Actions runs the repository's backend and test matrix

### DIFF
--- a/src/pyrecest/filters/abstract_particle_filter.py
+++ b/src/pyrecest/filters/abstract_particle_filter.py
@@ -6,6 +6,7 @@ from collections.abc import Callable
 from pyrecest.backend import (
     all,
     array,
+    asarray,
     hstack,
     isfinite,
     ndim,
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing because this branch is broadly red rather than near-mergeable: the test workflow fails across NumPy, JAX, and PyTorch jobs. The transactional particle-state change is valuable, but it needs a fresh implementation and backend-matrix validation rather than an unsafe merge during cleanup.

PR #4787, opened 2026-07-26
Promote Gaussian mixture mean shifts

FlorianPfaff:
## Summary

- shift Gaussian-mixture component means out of place so backend dtype promotion can occur
- preserve the original mixture while returning the translated copy
- add a regression covering integer-valued component means and a fractional target mean

## Bug

`GaussianMixture.set_mean()` deep-copied the mixture and then updated each component with `dist.mu += mean_offset`. When a valid Gaussian used an integer-valued mean array, the weighted mixture mean and requested shift were floating point. NumPy therefore rejected the in-place update with a casting error instead of returning the shifted distribution.

For example, component means `[0]` and `[2]` with weights `[0.25, 0.75]` cannot be shifted to mean `4.25`, even though the operation is mathematically valid.

## Fix

Assign `dist.mu + mean_offset` back to each copied component. This retains the intended translation and allows NumPy, PyTorch, and JAX to apply their normal numeric promotion rules.

## Validation

- reproduced the pre-fix NumPy `UFuncTypeError` for integer arrays shifted by floating-point offsets
- verified the out-of-place operation promotes the result to floating point
- added regression assertions for the target mean, both translated components, covariance preservation, and immutability of the original mixture
- branch is 2 commits ahead of current `main`, 0 behind; diff is limited to 2 files

### DIFF
--- a/src/pyrecest/distributions/nonperiodic/gaussian_mixture.py
+++ b/src/pyrecest/distributions/nonperiodic/gaussian_mixture.py
@@ -40,7 +40,7 @@ class GaussianMixture(LinearMixture, AbstractLinearDistribution):
         new_mixture = copy.deepcopy(self)
         mean_offset = new_mean - self.mean()
         for dist in new_mixture.dists:
-            dist.mu += mean_offset  # type: ignore
+            dist.mu = dist.mu + mean_offset  # type: ignore
         return new_mixture
 
     def to_gaussian(self, check_validity=True):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as backend-incomplete. NumPy and JAX pass, but the PyTorch matrix fails across supported Python versions. The intended integer-to-floating mean promotion needs an explicitly backend-safe implementation and PyTorch regression before it is suitable for merge.

PR #4800, opened 2026-07-26
Reject temporal hyperspherical sample counts

FlorianPfaff:
## Bug

`HypersphericalUniformDistribution.sample()` normalizes its sample count through a zero-dimensional NumPy array followed by `.item()`. For `np.timedelta64(4, "ns")`, NumPy returns the plain Python integer `4`, so a duration is silently accepted and four samples are generated. Other temporal units fail differently, making validation depend on the datetime unit.

## Fix

- Reject NumPy `datetime64` and `timedelta64` scalar dtypes before calling `.item()`.
- Preserve existing support for Python integers, NumPy integer scalars, and exact scalar numeric counts.
- Add regression coverage for nanosecond and microsecond timedeltas and a datetime scalar.

## Validation

- Standalone validator regression passes for accepted numeric counts and rejected temporal scalars.
- Branch comparison is limited to the hyperspherical sample-count validator and its focused test module.
- Branch is 2 commits ahead of `main` and 0 behind.

GitHub Actions provides the full backend test matrix.

### DIFF
--- a/src/pyrecest/distributions/hypersphere_subset/hyperspherical_uniform_distribution.py
+++ b/src/pyrecest/distributions/hypersphere_subset/hyperspherical_uniform_distribution.py
@@ -29,6 +29,8 @@ def _validate_positive_sample_count(n) -> int:
     count_array = np.asarray(n)
     if count_array.ndim != 0:
         raise ValueError("n must be a scalar integer")
+    if count_array.dtype.kind in ("m", "M"):
+        raise ValueError("n must be an integer")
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as redundant with #4803. Both add the same temporal-dtype guard to `_validate_positive_sample_count`, while #4803 places the regression alongside the validator's exactness tests and explicitly preserves the existing large exact-integer behavior. #4803 has no failed checks and will be retained for auto-merge.

PR #4859, opened 2026-07-27
Preserve sparse mode probabilities

FlorianPfaff:
## Bug

`ModeRBPFManifoldUKFTracker` normalized transition and initial-mode probabilities by flooring every entry at `1e-9`. This changed exact zero probabilities into small positive values, so modes and transitions that callers explicitly declared impossible became possible. The same flooring also silently converted negative entries into positive probability mass instead of rejecting malformed configurations.

## Fix

- preserve exact zero probabilities while normalizing valid rows and vectors;
- reject negative and non-finite transition probabilities;
- reject transition rows with zero total mass;
- validate initial-mode vector shape, finiteness, non-negativity, and positive total mass;
- add focused regressions for deterministic transitions and invalid probability specifications.

## Validation

The branch includes a one-shot workflow that applies the two-file patch, runs syntax checks and the focused mode-tracker test module, commits the validated result, and removes itself. The PR is draft while that workflow completes.


FlorianPfaff:
Closing this empty scaffold PR. The connector-created workflow could not self-apply because pull-request events only loaded workflows from the base branch. The actual bug fix is being published separately as a direct two-file change.

PR #4912, opened 2026-07-29
Reject masked and mixed-boolean model inputs

FlorianPfaff:
## What changed

- inspect raw model-validation inputs before NumPy/backend coercion
- reject masked vector and matrix entries instead of consuming their hidden payloads
- reject masked Boolean flags, dimensions, symmetry tolerances, and inferred dimensions
- reject Python sequences that mix Boolean and numeric entries before `True`/`False` are promoted to `1`/`0`
- preserve support for fully unmasked `numpy.ma.MaskedArray` inputs
- add focused regression coverage for array inputs, scalar metadata, dimension inference, and unmasked compatibility

## Root cause

`numpy.asarray` and backend conversion can discard a `MaskedArray` mask and expose its underlying data. They can also promote mixed numeric/Boolean Python sequences to a numeric dtype. The existing validation therefore ran after the information needed to reject those inputs had already been lost.

## Impact

Missing state, measurement, covariance, or model-matrix entries can no longer silently affect estimation through hidden values. Boolean configuration mistakes in numeric sequences also fail at the model boundary rather than changing values numerically.

## Validation

- reproduced the pre-fix acceptance of masked state entries, masked dimensions/flags/tolerances, masked inferred dimensions, and mixed Boolean/numeric arrays
- ran the focused regression module against the patched validation code with a NumPy-backed test harness: `7 passed`
- syntax-compiled the changed source and regression test

GitHub Actions provides the authoritative full repository and backend test matrix.

### DIFF
--- a/src/pyrecest/models/validation.py
+++ b/src/pyrecest/models/validation.py
@@ -20,8 +20,40 @@ _TEMPORAL_ARRAY_KINDS = {"M", "m"}
 _TEMPORAL_SCALAR_TYPES = (np.datetime64, np.timedelta64)
 
 
+def _contains_masked_value(value: Any) -> bool:
+    """Return whether ``value`` contains at least one masked entry."""
+    if np.ma.is_masked(value):
+        return True
+    if isinstance(value, np.ndarray):
+        if value.dtype == object:
+            return any(_contains_masked_value(item) for item in value.reshape(-1))
+        return False
+    if isinstance(value, (list, tuple)):
+        return any(_contains_masked_value(item) for item in value)
+    return False
+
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded and overly broad. The masked-input validation fix is covered more cleanly by #4915, including additive-noise vectorization flags and broader masked-input regression coverage. This PR additionally rejects mixed Boolean/numeric sequences globally, which expands API behavior and currently fails the test workflow; that behavior should be reconsidered separately only with a demonstrated downstream bug and compatibility analysis.

PR #4914, opened 2026-07-29
Fix JAX real FFT length array validation

FlorianPfaff:
## Summary
- require NumPy/JAX array-valued `n` arguments for `rfft` and `irfft` to be zero-dimensional integer scalars
- stop silently converting one-element vectors and matrices into scalar FFT lengths
- preserve support for genuine zero-dimensional NumPy and JAX integer arrays
- cover both real FFT directions and singleton/non-singleton nonscalar arrays

## Bug
`_normalize_real_fft_length` used `n.size == 1`, so values such as `np.array([3])`, `np.array([[3]])`, and their JAX equivalents were converted to `3`. NumPy's FFT contract rejects these nonscalar arrays with `TypeError`; PyRecEst's JAX wrapper instead accepted them and changed the transform length.

## Validation
- `PYTHONPATH=/mnt/data/pyrecest_fft_fix pytest -q /mnt/data/pyrecest_fft_fix/tests/test_jax_fft_scalar_array_axes.py`
- result: `3 passed`

### DIFF
--- a/src/pyrecest/_backend/jax/fft.py
+++ b/src/pyrecest/_backend/jax/fft.py
@@ -43,14 +43,14 @@ def _normalize_real_fft_length(n):
     if isinstance(n, _np.ndarray):
         if _np.issubdtype(n.dtype, _np.bool_):
             raise TypeError(_BOOLEAN_FFT_LENGTH_ERROR)
-        if n.size == 1 and _np.issubdtype(n.dtype, _np.integer):
+        if n.ndim == 0 and _np.issubdtype(n.dtype, _np.integer):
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as an incompatible API-tightening change rather than a correctness fix. The JAX backend already has an explicit cross-backend compatibility test that accepts singleton array-valued FFT lengths, and this PR makes that established test fail on all JAX versions. PyRecEst is not required to mirror NumPy's rejection semantics here; changing the contract would be a backward-incompatible decision that should be proposed separately, not merged as a bug fix.

PR #4951, opened 2026-07-30
Fix overflow in point-set distance norms

FlorianPfaff:
## What changed

- compute built-in OSPA/GOSPA Euclidean pairwise distances with a scale-normalized norm
- preserve finite distances when the naïve sum of squares would overflow
- add public OSPA and GOSPA regressions at large finite coordinates
- retain the existing metrics implementation byte-for-byte in `_metrics_impl.py`; the public `metrics.py` module re-exports that API after installing the corrected pairwise-distance kernel

## Root cause

`_pairwise_distances()` delegated directly to `np.linalg.norm`. For a finite difference vector such as `[1e308, 1e308]`, the squared intermediate overflows even though the Euclidean norm, approximately `1.4142e308`, is representable. OSPA/GOSPA then clip the resulting infinity to the cutoff and overstate localization error whenever the cutoff is above the true distance.

## Impact

Point-set metrics now retain the correct representable distance for extreme finite coordinates. Ordinary-scale behavior and custom `distance_fn` handling remain unchanged.

## Validation

- reproduced the pre-fix overflow under `np.errstate(over="raise", invalid="raise")`
- verified the corrected result against `np.hypot(1e308, 1e308)`
- verified ordinary `[3, 4]` and zero-distance cases
- focused regression tests cover both `ospa_distance()` and `gospa_distance()`
- repository CI is running on the final one-commit branch


FlorianPfaff:
Closing this PR because the implementation is disproportionate to the edge-case fix. It replaces the public 725-line metrics module with a runtime monkey-patching wrapper and duplicates the implementation in a private module, producing 1,500+ changed lines for a small numerical guard. CI is green, but this structure adds unnecessary maintenance and import-order risk. A future fix should update `_pairwise_distances` in place with a focused regression.

PR #4955, opened 2026-07-30
Stabilize retrodiction covariance symmetrization

FlorianPfaff:
## Summary
- replace overflow-prone covariance re-symmetrization in linear-Gaussian retrodiction with a backend-native stable average
- preserve extreme finite covariance entries, including same-sign values near `float64` maximum
- add a focused regression for identity retrodiction at the largest finite `float64` covariance scale

## Root cause
`retrodict_linear_gaussian` used `0.5 * (P + P.T)`. For finite entries near the maximum representable value, the intermediate sum can overflow even when the mathematical average is finite. This can turn a valid retrodicted covariance into `inf` or raise under strict floating-point error handling.

## Impact
Out-of-sequence retrodiction remains finite for representable extreme covariances. The replacement stays in the active numerical backend instead of converting through NumPy, preserving backend device and differentiation behavior.

## Validation
- standalone NumPy and PyTorch checks cover maximum finite values, minimum subnormals, and opposite-sign transpose pairs
- the one-shot workflow applies the actual source/test diff, validates syntax and arithmetic, removes itself, and commits the focused patch
- normal repository CI runs on the resulting source-and-test head


FlorianPfaff:
Closing this incomplete scaffold. The PR currently changes only `.github/workflows/one-shot-retrodiction-symmetrization.yml`; the self-patching workflow failed, so the advertised source fix and regression are not present in the diff. Please reopen or replace it with a focused PR containing the actual code/test changes and ordinary CI.

FlorianPfaff:
The requested source fix and regression are now present and the temporary workflow has been removed. I am replacing this PR because its head/check state remained tied to the earlier scaffold run; the replacement will contain the same focused two-file diff on the current branch head and ordinary CI.

PR #4961, opened 2026-07-31
Fix overflow in point-set distance norms

FlorianPfaff:
## What changed

- replace the overflow-prone built-in Euclidean pairwise norm with scale-normalized norm evaluation
- keep the correction directly inside `src/pyrecest/utils/metrics.py`
- add public OSPA and GOSPA regressions for large finite coordinates

## Root cause

`_pairwise_distances()` delegated directly to `np.linalg.norm`. For a finite difference vector such as `[1e308, 1e308]`, the internal sum of squares overflows even though the Euclidean norm, approximately `1.4142e308`, is representable. OSPA and GOSPA then clip the resulting infinity to the cutoff and overstate localization error whenever the cutoff exceeds the true distance.

## Impact

Built-in OSPA/GOSPA distances now preserve representable Euclidean distances at extreme finite coordinate scales. Ordinary-scale behavior and custom `distance_fn` handling are unchanged.

This replaces the valid bug fix attempted in #4951 with the requested focused implementation: no wrapper module, no duplicated metrics implementation, and no import-time monkey patching.

## Validation

- reproduced the pre-fix overflow under `np.errstate(over="raise", invalid="raise")`
- validated both public OSPA and GOSPA calls against `np.hypot(1e308, 1e308)`
- syntax-compiled the modified source and regression module
- branch is current with `main`; final diff is limited to one in-place source change and one regression file

GitHub Actions provides the full supported-version and backend validation matrix.


FlorianPfaff:
Closing as low value. This changes core OSPA/GOSPA distance evaluation solely for coordinate differences around 1e308. That scale is not a meaningful supported tracking regime, so the special normalization path and regression surface are not justified.

PR #4964, opened 2026-07-31
Reject masked shared random parameters

FlorianPfaff:
## Summary

- reject genuinely masked distribution parameters before shared NumPy/Autograd random wrappers coerce them with `asarray`
- cover uniform bounds, normal location/scale, multivariate-normal mean/covariance, and `choice` probabilities
- preserve ordinary NumPy inputs and fully unmasked `MaskedArray` values

## Root cause

The shared validators converted inputs with `asarray` before checking NumPy mask semantics. For a masked scalar or partially masked array, that conversion exposes the hidden data payload and discards the mask. Missing configuration values could therefore become real sampling parameters.

Examples include a masked normal scale being silently used as its hidden standard deviation and a masked `choice` probability being normalized and sampled from.

## Impact

NumPy and Autograd backend samplers now fail early instead of drawing reproducible-looking samples from missing data. The earlier masked-input fix for NumPy `randint` and `multinomial` did not cover these shared distribution samplers.

## Validation

- reproduced mask loss for scalar and partially masked arrays through NumPy `asarray`
- isolated exact-module regression: eight masked cases rejected and the fully unmasked compatibility path passed
- Python syntax compilation passed for the modified source and regression module
- branch comparison: 2 commits ahead, 0 behind `main`
- final diff is limited to 8 validator lines and one focused test file

### DIFF
--- a/src/pyrecest/_backend/_shared_numpy/random.py
+++ b/src/pyrecest/_backend/_shared_numpy/random.py
@@ -84,6 +84,8 @@ def _normalize_size(size):
 
 
 def _validate_uniform_bound(bound, name):
+    if _contains_masked_value(bound):
+        raise TypeError(f"{name} must be real numeric")
     if _contains_boolean_value(bound):
         raise TypeError(f"{name} must be real numeric, not boolean")
     try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This expands masked-array rejection across shared random-distribution parameters, but masked arrays are not a supported cross-backend sampling contract in PyRecEst. Adding piecemeal validation here would increase inconsistency rather than establish coherent missing-data behavior.

PR #4966, opened 2026-07-31
Reject masked JAX matrix-power exponents

FlorianPfaff:
## Bug

The JAX linalg backend normalized `matrix_power` exponents with `operator.index` before checking NumPy masked-array semantics. A masked scalar such as `np.ma.array(2, mask=True)` therefore exposed its hidden payload and was accepted as exponent `2`.

This can turn a missing configuration value into a real matrix operation instead of failing deterministically.

## Fix

- reject genuinely masked exponent inputs before integer coercion;
- preserve ordinary Python, NumPy, and JAX integer scalar handling;
- preserve fully unmasked integer-valued `MaskedArray` scalars;
- add focused subprocess regression coverage on the JAX backend.

## Validation

- reproduced that `operator.index(np.ma.array(2, mask=True))` returns the hidden value `2`;
- focused local JAX validation passed for two masked rejection cases and the unmasked compatibility case;
- branch comparison: **2 commits ahead, 0 behind** `main`;
- final diff is limited to **2 implementation lines and 21 test lines** across two files.

GitHub Actions provides the full supported JAX, Python-version, lint, and packaging matrix.

### DIFF
--- a/src/pyrecest/_backend/jax/linalg.py
+++ b/src/pyrecest/_backend/jax/linalg.py
@@ -18,6 +18,8 @@ def _as_linalg_array(value):
 
 def _as_integer_scalar(value, name):
     """Return a Python integer for JAX static integer arguments."""
+    if _np.ma.is_masked(value):
+        raise TypeError(f"{name} must be an integer scalar")
     if isinstance(value, (bool, _np.bool_)):
         raise TypeError(f"{name} must be an integer scalar")
     try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This only adds rejection semantics for NumPy masked scalars passed as JAX `matrix_power` exponents. Masked arrays are outside the normal JAX/PyRecEst parameter contract, so isolated support would add backend-specific complexity without practical benefit.

PR #4967, opened 2026-07-31
Honor masks in PyTorch nonzero

FlorianPfaff:
## Bug

The PyTorch backend wrapper converted NumPy `MaskedArray` inputs with `torch.as_tensor`. That conversion discards the mask and exposes the hidden payload, so `nonzero` could report masked entries as present.

For example, a masked 2×2 input whose only unmasked nonzero value is at `(1, 0)` returned `(0, 1)`, `(1, 0)`, and `(1, 1)` on the PyTorch backend.

## Fix

- fill masked entries with zero before converting to a PyTorch tensor;
- preserve the existing rejection of zero-dimensional inputs;
- preserve ordinary list, NumPy-array, and tensor handling;
- add regression coverage for both the public and raw PyTorch backend helpers.

## Validation

- reproduced the pre-fix mask loss with NumPy 2.3.5 and PyTorch 2.10.0;
- verified the patched helper matches `numpy.nonzero` for the masked input;
- verified existing scalar rejection and ordinary matrix behavior in the isolated harness;
- syntax-compiled both modified files;
- branch comparison: 2 commits ahead, 0 behind `main`, with only 2 files changed.

The full supported backend and Python-version matrix is delegated to GitHub Actions.

### DIFF
--- a/src/pyrecest/backend_support/_pytorch_nonzero_scalar_contract.py
+++ b/src/pyrecest/backend_support/_pytorch_nonzero_scalar_contract.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import numpy as np
+
 _NONZERO_SCALAR_MESSAGE = (
     "Calling nonzero on 0d arrays is not allowed. "
     "Use np.atleast_1d(scalar).nonzero() instead."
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This introduces NumPy masked-array semantics into the PyTorch `nonzero` compatibility layer, while masked arrays are not a native PyTorch input contract and are not supported consistently across PyRecEst. Isolated mask handling here would make backend behavior less coherent.

PR #4971, opened 2026-08-01
Stabilize time-offset metrics at extreme scales

FlorianPfaff:
## Summary

- compute calibration residual norms with `numpy.hypot.reduce`
- evaluate mean, standard deviation, RMSE, and count-weighted aggregates after scale normalization
- pool standard deviations in centered, scaled coordinates
- add strict-overflow regressions for representable `float64` residuals and summaries

## Root cause

`time_offset_error_summary()` used `numpy.linalg.norm`, while its summary and aggregation helpers squared or directly reduced values near the `float64` limit. These intermediate operations can overflow even when the final metric remains representable.

For example, a residual `[1e308, 1e308]` has the finite Euclidean norm `1.4142135623730951e308`, but the current path overflows while squaring. The resulting non-finite residual is then discarded, reducing the reported count and coverage. Count-weighted RMSE aggregation has the same issue for a finite `1e308` summary value.

## Fix

- use scale-aware Euclidean norm accumulation for residual rows;
- normalize values before mean, standard-deviation, and RMS calculations;
- normalize weights before weighted reductions;
- compute pooled variance from centered scaled group moments instead of subtracting two large second moments.

## Validation

- reproduced both failures on current `main` with `numpy.errstate(over="raise", invalid="raise")`;
- pre-fix focused regressions: **2 failed**;
- patched focused regressions: **2 passed**;
- syntax compilation passed for the modified source and test;
- the branch-local workflow installs the package, runs the focused regression, then removes itself and the patch helper so the final diff contains only the source fix and regression test.


FlorianPfaff:
Closing as low value. The substantial new stable-statistics machinery is dedicated to calibration residuals and summaries at near-maximum float64 magnitudes (around 1e308), outside a meaningful supported calibration regime. The complexity and maintenance cost outweigh practical benefit.

PR #4972, opened 2026-08-01
Reject masked JAX FFT parameters

FlorianPfaff:
## Bug

The JAX FFT compatibility layer accepts NumPy `MaskedArray` scalar parameters by calling `.item()` or `operator.index()` on them. Those operations expose the hidden payload even when the scalar is masked.

For example, `n=np.ma.array(4, mask=True)` is silently normalized to `4`, so `rfft` executes a real transform instead of rejecting a missing length. The same issue affects real-FFT axes, scalar entries inside `fftn`/`ifftn` shape and axis sequences, and FFT-shift axes.

## Fix

- detect NumPy masked arrays containing masked values before integer normalization;
- reject masked real-FFT lengths and axes with parameter-specific `TypeError`s;
- reject masked entries in complex FFT shape/axis sequences;
- reject masked FFT-shift axes;
- preserve support for fully unmasked `MaskedArray` integer parameters.

## Impact

Missing FFT configuration can no longer become a reproducible-looking transform through the masked value's hidden payload. Ordinary Python, NumPy, and JAX integer parameters retain their existing behavior.

## Validation

- reproduced the pre-fix behavior with NumPy 2.3.5 and JAX 0.9.0.1: a masked `n=4` was accepted and executed;
- exercised the patched normalization for masked `n`, real-FFT `axis`, complex FFT `s`/`axes` entries, and shift axes;
- verified fully unmasked `MaskedArray` integer parameters remain accepted;
- syntax-compiled the isolated patched module;
- branch comparison is limited to the JAX FFT wrapper and one focused regression file.

GitHub Actions will provide the repository's full supported Python, backend, lint, and packaging validation matrix.

### DIFF
--- a/src/pyrecest/_backend/jax/fft.py
+++ b/src/pyrecest/_backend/jax/fft.py
@@ -8,13 +8,24 @@ from jax.numpy import fft as _fft
 
 _BOOLEAN_FFT_AXIS_ERROR = "axis must be an integer, not boolean"
 _BOOLEAN_FFT_LENGTH_ERROR = "n must be an integer length, not boolean"
+_FFT_AXIS_ERROR = "axis must be an integer"
+_FFT_LENGTH_ERROR = "n must be an integer length"
 _FFT_SHAPE_SEQUENCE_ERROR = "s must be None or a sequence of integer lengths"
 _FFT_AXES_SEQUENCE_ERROR = "axes must be None or a sequence of integer axes"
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This adds masked-array semantics to JAX FFT configuration parameters, but masked arrays are not part of the normal JAX/PyRecEst FFT contract. Supporting them here in isolation would create inconsistent behavior without broader end-to-end masked-array support.

PR #4974, opened 2026-08-01
Fix overflow in tracklet edge distances

FlorianPfaff:
## Bug

`constant_velocity_edge_cost()` computed endpoint displacement with `numpy.linalg.norm`. For large but finite coordinates such as `[1e308, 1e308]`, the internal sum of squares overflows even though the Euclidean norm (about `1.4142e308`) is still representable.

The resulting infinite edge cost is filtered out by `build_tracklet_adjacency()`, so a mathematically valid and feasible tracklet transition can disappear from the association graph solely because of an avoidable intermediate overflow.

## Fix

- compute the Euclidean displacement with NumPy's scale-safe repeated `hypot` reduction
- preserve ordinary-scale edge costs and speed gating
- add a regression under strict overflow/invalid floating-point handling at the largest useful finite scale

## Validation

- reproduced the current failure locally: `numpy.linalg.norm([1e308, 1e308])` overflows to infinity
- verified the replacement returns the finite value from `numpy.hypot(1e308, 1e308)` and preserves a `[3, 4]` displacement cost of `5`
- the branch-local focused validation syntax-compiles the modified source and regression before committing the final two-file patch

The repository's normal pull-request checks provide the full integration matrix.


FlorianPfaff:
Closing as low value. This hardens tracklet rotation only for coordinates near the floating-point limit (around 1e308), which is not a practical supported state-estimation regime. The added scale-specific machinery would increase complexity without meaningful application benefit.

PR #4975, opened 2026-08-01
Stabilize SO(3) product quaternion normalization

FlorianPfaff:
## Bug

`SO3ProductDiracDistribution` normalized quaternions with a direct backend Euclidean norm. For large but finite values such as `[1e308, 1e308, 0, 0]`, the sum-of-squares intermediate overflows even though the true norm, approximately `1.4142e308`, is representable.

The constructor, geodesic-distance helper, and rotation-matrix conversion could therefore reject a finite, nonzero quaternion solely because of avoidable intermediate overflow.

## Fix

- reuse the existing scale-safe `normalize_quaternions()` helper shared by PyRecEst's other SO(3) distributions;
- retain finite/nonzero validation and canonical scalar-last quaternion handling;
- add a regression that constructs an SO(3) product Dirac distribution from extreme finite coordinates under strict overflow/invalid handling.

## Validation

- reproduced the old failure locally: `numpy.linalg.norm([1e308, 1e308, 0, 0])` raises `FloatingPointError` with overflow configured to raise;
- verified the scale-safe normalization returns `[sqrt(0.5), sqrt(0.5), 0, 0]` with unit norm;
- syntax-compiled the modified source and regression test;
- branch is 2 commits ahead and 0 behind `main`, with exactly two changed files.

The repository's pull-request workflows provide the full supported backend, Python-version, lint, and packaging validation matrix.

### DIFF
--- a/src/pyrecest/distributions/so3_product_dirac_distribution.py
+++ b/src/pyrecest/distributions/so3_product_dirac_distribution.py
@@ -25,6 +25,7 @@ from pyrecest.backend import (
     where,
 )
 
+from ._so3_helpers import normalize_quaternions as _normalize_so3_quaternions
 from .cart_prod.hyperhemisphere_cart_prod_dirac_distribution import (
     HyperhemisphereCartProdDiracDistribution,
 )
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing because it is not merge-ready. The full test workflow fails broadly across the supported backend/Python matrix, despite the focused change looking reasonable. The normalization-helper reuse can be reconsidered as a smaller PR once it has a fully green matrix.

PR #4977, opened 2026-08-03
Reject out-of-range index-matrix fill values

FlorianPfaff:
## Bug

`tracks_to_index_matrix()` validates that `fill_value` is a negative integer, but did not validate that it is representable by the integer matrix it allocates. A value below `numpy.iinfo(numpy.int64).min` therefore passed the public validation and failed later inside the backend allocator with a raw `OverflowError`.

For example, `fill_value=numpy.iinfo(numpy.int64).min - 1` reached `numpy.full(..., dtype=int)` and raised `OverflowError: Python int too large to convert to C long`.

## Fix

- reject negative fill values below the `int64` range with a deterministic `ValueError`;
- preserve the valid `int64` minimum boundary;
- add regression coverage for both the immediate underflow case and a much larger negative Python integer.

## Validation

- reproduced the pre-fix `OverflowError` through the current module in an isolated NumPy-backend harness;
- the patched focused test module passes all three index-matrix tests;
- syntax compilation passes for both modified files;
- branch comparison is limited to two files: 3 source lines and 23 regression-test lines.

The repository's pull-request workflows provide the full supported Python/backend, lint, packaging, documentation, and security validation matrix.

### DIFF
--- a/src/pyrecest/utils/multisession_assignment_score.py
+++ b/src/pyrecest/utils/multisession_assignment_score.py
@@ -50,6 +50,7 @@ _INVALID_SCORE_SCALAR_TYPES = (
 )
 _REJECTED_SCORE_ARRAY_KINDS = frozenset({"b", "c", "S", "U", "M", "m"})
 _SCORE_TO_COST_ERROR = "score_to_cost must return real numeric cost matrices."
+_INT64_MIN = int(np.iinfo(np.int64).min)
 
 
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This only changes the exception produced for an already-invalid integer sentinel outside the representable int64 range. Supported sentinel values are unaffected, so the maintenance and test surface is not justified by practical user impact.

PR #4980, opened 2026-08-03
Reject masked assignment inputs

FlorianPfaff:
## What changed

- detect genuinely masked NumPy values before backend conversion
- reject masked assignment counts, cost matrices, and row/column non-assignment costs with explicit validation errors
- preserve support for fully unmasked `MaskedArray` inputs
- add regression coverage for both assignment solvers and nested object-array inputs

## Root cause

Backend/NumPy array conversion can discard `MaskedArray` mask metadata and expose the hidden payload. In the assignment utilities, a missing cost could therefore be interpreted as a real numeric cost and silently change the selected assignment; a masked `k` could likewise be converted into an unintended count.

## Validation

- reproduced the pre-fix behavior: masked matrix entries were accepted and selected using their hidden values
- focused regression suite: 4 tests passed
- source and test files compile successfully
- GitHub CI should run the repository's full supported backend/test matrix on this PR

### DIFF
--- a/src/pyrecest/utils/assignment.py
+++ b/src/pyrecest/utils/assignment.py
@@ -33,6 +33,31 @@ _TEMPORAL_TYPES = (_np.datetime64, _np.timedelta64)
 _INVALID_SCALAR_TYPES = _BOOLEAN_TYPES + _TEXT_TYPES + _TEMPORAL_TYPES
 
 
+def _contains_masked_values(value, active_ids: set[int] | None = None) -> bool:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. Recursively propagating NumPy masked-array semantics through the assignment solvers adds substantial complexity for an input type that is not otherwise supported consistently by the solver stack. A broader API-level design would be needed before adopting this behavior.

PR #4981, opened 2026-08-03
Stabilize scalar DVS normal-flow normalization

FlorianPfaff:
## Bug

`pyrecest.experimental.dvs.signed_normal_flow()` normalized contour normals and velocities with `numpy.linalg.norm`. For large but finite vectors such as `[1e308, 1e308]`, the internal sum of squares overflows even though the true Euclidean norm remains representable.

For a horizontal normal and a 45-degree velocity at that scale, the helper returned `0.0` instead of `sqrt(0.5)`. With NumPy overflow handling configured to raise, it raised `FloatingPointError`. The profile and activity helpers inherit the same failure because they call this scalar path.

The vectorized DVS tracker path already uses scale-safe velocity normalization, but the public scalar active-contour helpers remained vulnerable.

## Fix

- compute Euclidean norms with a scale-safe `numpy.hypot.reduce` helper;
- normalize both the normal and velocity before taking their dot product, keeping the final arithmetic bounded;
- preserve the existing zero-vector behavior;
- add focused scalar and profile regressions using extreme finite normals and velocity.

## Validation

- reproduced the pre-fix result: `0.0` instead of `0.7071067811865475`;
- reproduced the pre-fix `FloatingPointError` under `numpy.errstate(over="raise", invalid="raise")`;
- verified the patched scalar/profile calculation returns `[sqrt(0.5), sqrt(0.5), -sqrt(0.5)]` without overflow;
- branch is 2 commits ahead and 0 behind `main`;
- final diff is limited to the active-contour helper and its focused test module.

GitHub Actions should provide the full supported Python, backend, lint, packaging, documentation, and security validation matrix.

### DIFF
--- a/src/pyrecest/experimental/dvs/active_contour.py
+++ b/src/pyrecest/experimental/dvs/active_contour.py
@@ -21,6 +21,14 @@ def unit_vector_from_angle(angle: float) -> np.ndarray:
     return np.array([np.cos(angle), np.sin(angle)], dtype=float)
 
 
+def _stable_euclidean_norm(vector: np.ndarray) -> float:
+    """Return a scale-safe Euclidean norm for a finite vector."""
+    absolute_values = np.abs(vector).reshape(-1)
+    if absolute_values.size == 0:
+        return 0.0
+    return float(np.hypot.reduce(absolute_values))
+
+
 def signed_normal_flow(normal: np.ndarray, velocity: np.ndarray) -> float:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. The change is dedicated to normalization at near-maximum floating-point magnitudes (around 1e308), where the downstream geometry is not a meaningful supported operating regime. The added special-case path would increase maintenance complexity without practical benefit.

PR #4983, opened 2026-08-03
Reject masked JAX norm axes

FlorianPfaff:
## What changed

- reject genuinely masked NumPy axis values before JAX/static-axis coercion
- cover direct masked axes and masked scalar entries nested in lists or tuples
- preserve valid ordinary static-axis inputs and fully unmasked nested scalar axes
- add a focused backend-portable regression

## Root cause

The JAX `linalg.norm` compatibility layer normalizes list and tuple axes entry by entry. For a masked NumPy scalar such as `np.ma.array(1, mask=True)`, `operator.index()` exposes the hidden payload. Consequently, `axis=[np.ma.array(1, mask=True)]` was silently normalized to `(1,)` and the norm was evaluated along axis 1 instead of rejecting a missing axis value.

Direct masked arrays could also leak backend-specific `ValueError`s during JAX conversion rather than producing the documented axis validation error.

## Impact

Missing or masked configuration can no longer become a real norm axis through its hidden payload. Callers now receive a deterministic `TypeError` for all genuinely masked axis forms.

## Validation

- reproduced the pre-fix nested-axis behavior: a masked scalar payload of `1` produced `[5.0, 13.0]`
- exercised the patched normalizer against scalar, vector, nested-list, and nested-tuple masked axes
- verified all masked cases raise the documented `TypeError`
- verified existing NumPy/JAX scalar, array, list, and tuple axis forms still produce `[5.0, 13.0]`
- final comparison is 2 commits ahead and 0 behind `main`, limited to 4 source lines and one focused regression file

GitHub Actions will run the repository's full supported backend and Python-version matrix.

### DIFF
--- a/src/pyrecest/_backend/jax/linalg.py
+++ b/src/pyrecest/_backend/jax/linalg.py
@@ -35,6 +35,8 @@ def _as_integer_scalar(value, name):
 
 def _norm_axis_entry(axis) -> int:
     """Return one non-boolean NumPy-style linalg.norm axis."""
+    if _np.ma.is_masked(axis):
+        raise TypeError(_AXIS_TYPE_ERROR)
     if isinstance(axis, (bool, _np.bool_)):
         raise TypeError(_AXIS_TYPE_ERROR)
     try:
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as low value. This adds masked-array semantics to a JAX batched-solve path, but masked arrays are not part of the normal JAX/PyRecEst array contract. Supporting them here in isolation would create an inconsistent API without broader end-to-end masked-array support.

